PR #1, opened 2023-12-14
fix!: compile errors, import paths and tests

cmwaters:
This PR does an initial sweep of the repo, defining the new go mod and dependencies. It ensures that the library compiles and all tests pass.

It tries to minimize breakage from the celestia-app repo


pkg/shares/sparse_shares_test.go
```diff
@@ -84,3 +82,17 @@ func TestSparseShareSplitterCount(t *testing.T) {
 		})
 	}
 }
+
+// GenerateRandomBlobOfShareCount returns a blob that spans the given
+// number of shares
```
rootulp, line 87:
```suggestion
// generateRandomBlobOfShareCount returns a blob that spans the given
// number of shares
```

pkg/square/builder.go
```diff
@@ -6,24 +6,22 @@ import (
...
+	"github.com/celestiaorg/go-square/pkg/namespace"
+	"github.com/celestiaorg/go-square/pkg/shares"
+	"google.golang.org/protobuf/proto"
 )
 
 type Builder struct {
-	// maxCapacity is an upper bound on the total amount of shares that could fit in the biggest square
-	maxCapacity int
+	// maxSquareSize is an upper bound on the total amount of shares that could fit in the biggest square
+	maxSquareSize int
```
rootulp, line 18: [blocking] I think this comment is now incorrect. Perhaps something like:

```suggestion
	// maxSquareSize is the maximum number of rows (or columns) in the original data square
	maxSquareSize int
```

pkg/square/builder_test.go
```diff
@@ -2,253 +2,152 @@ package square_test
...
+// 			rand := tmrand.NewRand()
+// 			signer, err := testnode.NewOfflineSigner()
+// 			require.NoError(t, err)
+// 			txs := generateMixedTxs(signer, rand, tt.normalTxs, tt.pfbCount, 1, tt.pfbSize)
+// 			square, _, err := square.Build(txs, appconsts.LatestVersion, appconsts.DefaultGovMaxSquareSize)
+// 			require.NoError(t, err)
+// 			require.EqualValues(t, tt.expectedSquareSize, square.Size())
+// 		})
+// 	}
+// }
```
rootulp, line 45: [blocking] is this intended to be committed commented out? Instead of merging commented code can we:
1. delete if no longer relevant
1. uncomment and fix 
2. worst case: delete with a GH issue to re-add?

pkg/square/builder_test.go
```diff
@@ -2,253 +2,152 @@ package square_test
...
-			require.NoError(t, err)
-			txs := generateBlobTxsWithNamespaces(t, ns1.Repeat(len(tc.blobSize)), [][]int{tc.blobSize})
-			require.Len(t, txs, 1)
-			blobTx, isBlobTx := blob.UnmarshalBlobTx(txs[0])
-			require.True(t, isBlobTx)
-			require.Equal(t, tc.added, builder.AppendBlobTx(blobTx))
-		})
-	}
-}
+// func TestBuilderRejectsBlobTransactions(t *testing.T) {
```
rootulp, line 55: [blocking] is this intended to be committed commented out? Instead of merging commented code can we:
1. delete if no longer relevant
1. uncomment and fix 
2. worst case: delete with a GH issue to re-add?

pkg/square/builder_test.go
```diff
@@ -2,253 +2,152 @@ package square_test
...
-		}
-		require.LessOrEqual(t, shareRange.End, size)
-		txShares := dataSquare[shareRange.Start : shareRange.End+1]
-		parsedShares, err := rawData(txShares)
-		require.NoError(t, err)
-		require.True(t, bytes.Contains(parsedShares, tx))
-		lastEnd = shareRange.End
-	}
-}
+// func TestBuilderFindTxShareRange(t *testing.T) {
```
rootulp, line 114: [blocking] is this intended to be committed commented out? Instead of merging commented code can we:

delete if no longer relevant
uncomment and fix
worst case: delete with a GH issue to re-add?

pkg/square/builder_test.go
```diff
@@ -262,178 +161,186 @@ func rawData(shares []shares.Share) ([]byte, error) {
...
-				wrappedPFB, isWrappedPFB := coretypes.UnmarshalIndexWrapper(tx)
-				require.True(t, isWrappedPFB)
-				require.Equal(t, tt.expectedIndexes[j], wrappedPFB.ShareIndexes, j)
-			}
-		})
-	}
-}
+// // TestSquareBlobPositions ensures that the share commitment rules which dictate the padding
+// // between blobs is followed as well as the ordering of blobs by namespace.
+// func TestSquareBlobPostions(t *testing.T) {
```
rootulp, line 166: [blocking] is this intended to be committed commented out? Instead of merging commented code can we:

delete if no longer relevant
uncomment and fix
worst case: delete with a GH issue to re-add?

pkg/square/square.go
```diff
@@ -200,7 +188,7 @@ func (s Square) Size() int {
 // avoid breaking the api. In future versions there will not be a copy of this
 // code here.
 func Size(len int) int {
```
rootulp, line 190: [blocking] the godoc here needs to be updated to remove this portion:

"This function is currently a wrapper around the da packages equivalent function to avoid breaking the api."

Sorry can't provide a suggestion because lines not modified.

pkg/square/square_benchmark_test.go
```diff
@@ -1,56 +1,44 @@
 package square_test
 
-import (
-	"fmt"
-	"testing"
+// func BenchmarkSquareConstruct(b *testing.B) {
```
rootulp, line 3: [blocking] is this intended to be committed commented out? Instead of merging commented code can we:

delete if no longer relevant
uncomment and fix
worst case: delete with a GH issue to re-add?

proto/celestia/core/v1/blob/blob.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
-package celestia.core.v1.blob;
+package square.blob.v1;
```
rootulp, line 2: [question] Since this package name was modified, does the filename also need to be updated to remove `celestia/core`?

rootulp (commented):
Thanks for extracting this! I think this will lead to:
1. Fewer copies of this code floating around in other repos b/c they previously couldn't import from celestia-app
2. Ability to make breaking changes to the libs in this repo at a faster rate than when they were in celestia-app repo

Only blocking on commented tests. Not blocking but a few of the Godocs need to be updated.

proto/celestia/core/v1/blob/blob.proto
cmwaters: I was planning on modify the directory structure in a follow up PR

Wondertan:
Thank you for doing this! I am coming up with a suggestion. After playing with these types for years, I am convinced that keeping the core types Blob, Share, Namespace, and Square(EDS alias) under a single pkg would benefit devx, i.e. easier to import and make changes within the typ definition and for the user. This doesn't however mean that everything else should be in the same pkg. Something to consider for future PRs.

proto/celestia/core/v1/blob/blob.proto
rootulp: Works for me

pkg/inclusion/blob_share_commitment_rules.go
```diff
@@ -50,7 +50,7 @@ func BlobSharesUsedNonInteractiveDefaults(cursor, subtreeRootThreshold int, blob
 // fit in the square. The cursor is expected to be the index after the end of
 // the previous blob.
 //
-// See https://github.com/celestiaorg/celestia-app/blob/main/specs/src/specs/data_square_layout.md
+// See https://github.com/celestiaorg/go-square/blob/main/specs/src/specs/data_square_layout.md
```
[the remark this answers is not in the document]
cmwaters: good catch. yeah alternatively we go along with the every repo has it's own specs in which case the data_square_layout stuff should be eventually moved here

cmwaters (commented):
@rootulp in regards to the few commented out tests. I left them commented out instead of deleting them as more of a convenience for when I go back and re-enable the tests (which will require porting over a lot of the test factory code). I can delete them and use git diff to readd the tests

rootulp:
> @rootulp in regards to the few commented out tests. I left them commented out instead of deleting them as more of a convenience for when I go back and re-enable the tests (which will require porting over a lot of the test factory code). I can delete them and use git diff to readd the tests

I see. Honestly since this repo isn't used yet and doesn't have an official release, I'm okay with merging the commented code and uncommenting in a follow-up PR.

rootulp (approved):

cmwaters:
> I am convinced that keeping the core types Blob, Share, Namespace, and Square(EDS alias) under a single pkg would benefit devx

It should be easy just to add the types as aliases in the main package so you only need to import the main package to use all these types

cmwaters:
> I see. Honestly since this repo isn't used yet and doesn't have an official release, I'm okay with merging the commented code and uncommenting in a follow-up PR.

Going to merge this now and follow up with the leftover tests

PR #2, opened 2023-12-19
feat: add merkle package from cometbft

cmwaters:
This PR transfers the commits around the crypto/merkle package in our celestia-core fork to this repo.

It's defined as a separate go mod. We could do this as a separate repo but for now it's just go-square that uses it.

This is needed to enable the tests.

There are no changes to the library itself - just the import paths. If you want to assess the diff that I have made refer to the last commit: [init go mod, copy proto files and modify directory path](https://github.com/celestiaorg/go-square/pull/2/commits/fdf1e920e58a838598c959315a5c50b7f0ee54ce)


merkle/tree_test.go
```diff
@@ -159,3 +157,32 @@ func Test_getSplitPoint(t *testing.T) {
...
+	// Try a random mutation
+	switch rand.Int() % 2 {
+	case 0: // Mutate a single byte
+		bytez[rand.Int()%len(bytez)] += byte(rand.Int()%255 + 1)
+	case 1: // Remove an arbitrary byte
+		pos := rand.Int() % len(bytez)
+		bytez = append(bytez[:pos], bytez[pos+1:]...)
+	}
+	return bytez
+}
```
rootulp, line 188: note for reviewers, copied from https://github.com/celestiaorg/celestia-core/blob/34b83b51889c92a9818221bb01dc78a35dc55c77/libs/test/mutate.go#L7-L28

rootulp (approved):

PR #3, opened 2023-12-20
chore: remove unnecessary conversions

estensen:


rootulp (commented):
Overall LGTM. Before we merge this, I think we should merge a prerequisite PR that adds golangci-lint support.

Then this PR can add the unconvert linter to the golangci-lint config.

cmwaters (approved):
Thanks for the contribution

Issue #7, opened 2024-01-03
Add readme


cmwaters:
To explain the purpose and scope of the repo

PR #14, opened 2024-01-09
docs: README

rootulp:
Closes https://github.com/celestiaorg/go-square/issues/7

This is a barebones README. I plan on adding a **Helpful Commands** section after we add a Makefile.


cmwaters (approved):
Looks good as a starting point.

I think it would nice to add some small diagrams and an explanation of what this repo does and how to use it but we can look at that in a follow up PR

Issue #13, opened 2024-01-09
Add command to generate Go types from `.proto` files


rootulp:
## Problem

This repo has a `.proto` file but no command to generate the Go types from that `.proto` file.

## Proposal

1. Create a Makefile
2. Add a proto-gen command inspired by celestia-app

https://github.com/celestiaorg/celestia-app/blob/ebd733d1db2e75d849a637f68587534447ceccd0/Makefile#L55-L59

rootulp:
Hmm started this by copying files from celestia-app but encountered:

```
$ make proto-gen
--> Generating Protobuf files
/opt/homebrew/bin/docker run --rm -v /Users/rootulp/git/rootulp/celestiaorg/go-square:/workspace --workdir /workspace tendermintdev/sdk-proto-gen:v0.7 sh ./scripts/protocgen.sh
Failure: could not unmarshal as YAML: yaml: unmarshal errors:
  line 3: field plugin not found in type bufgen.ExternalPluginConfigV1
make: *** [proto-gen] Error 1
```

which seems caused by the `buf.gen.yaml` in this repo diverging from the `bug.gen.gogo.yaml` in celestia-app. I'm assuming that was intentional

cmwaters:
Yeah good point. I forgot that I generated the protos in a custom way and that we should add some command. I can look into this since I caused this mess

PR #16, opened 2024-01-10
fix: add commands for generating go types

cmwaters:
Closes: #13


buf.yaml
```diff
@@ -1,9 +1,9 @@
...
-  -proto
+name: buf.build/celestiaorg/go-square
 breaking:
   use:
     - FILE
 lint:
   use:
-    - DEFAULT
+    - BASIC
+
```
rootulp, line 9: to satisfy yamllint

```suggestion
```

Makefile
```diff
@@ -0,0 +1,20 @@
...
+proto-gen-docker:
+	@echo "--> Generating Protobuf files"
+	$(DOCKER_BUF) generate
+.PHONY: proto-gen-docker
+
+## proto-gen: Generate protobuf files. Requires buf.
+proto-gen:
+	@echo "--> Generating Protobuf files"
+	@buf generate
+.PHONY: proto-gen
```
rootulp, line 14: [question] do we want to support two different ways to generate the Protobuf files? I think it easier to maintain just one option and if that option is used in CI as well as locally. Since the `make proto-gen` command depends on a locally installed Buf CLI, preference for the Docker approach.

Ideally the Docker approach uses the same version of Buf locally and in CI.

rootulp (commented):
Note: I'm observing that when I run  `make proto-gen-docker` or `make proto-gen`, there are changes to these files:

```
	modified:   merkle/proto/proof.pb.go
	modified:   pkg/blob/blob.pb.go
```

staheri14 (approved):
LGTM!

Makefile
cmwaters: We could probably look to enforce it somehow. The docker image also isn't versioned (I actually had a problem where it wasn't working because I had an old image.

I think some people don't like using docker, hence why I added the other option, but it's simple enough that it probably doesn't need a make command so I will remove it

cmwaters:
> Note: I'm observing that when I run make proto-gen-docker or make proto-gen, there are changes to these files:

We might have different versions - let me dig into it

Makefile
rootulp: Yea I think the Docker approach should also be versioned.

rootulp:
If it helps debug:

```
$ buf --version
1.28.1
```

```
docker images

bufbuild/buf                                  latest    5085a683a6ba   3 months ago    55.4MB
```

staheri14:
> Note: I'm observing that when I run `make proto-gen-docker` or `make proto-gen`, there are changes to these files:
> 
> ```
> 	modified:   merkle/proto/proof.pb.go
> 	modified:   pkg/blob/blob.pb.go
> ```

It is the case for me too.

cmwaters:
Okay I had an older version of buf. I'm going to hardcode the version to 1.28.1 in the Makefile so hopefully that produces constant results

rootulp (approved):
`make proto-gen` results in no changes and `make proto-lint` results in no issues. Thanks!

rootulp:
We likely need a maintainer to make it so 2 approvals are needed in this repo.

rootulp:
Hmm we need to update settings on this repo so that new commits dismiss previous PR approvals.

cmwaters:
> Hmm we need to update settings on this repo so that new commits dismiss previous PR approvals.

Done. I've also made it require two reviews like the other repos

staheri14 (approved):

PR #29, opened 2024-01-24
chore(makefile): add lint, test, &  benchmark commands.

00x-dx:
Update Makefile with lint, test and benchmark commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,34 +2,38 @@ DOCKER := $(shell which docker)
 DOCKER_BUF := $(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace bufbuild/buf:1.28.1
 PROJECTNAME=$(shell basename "$(PWD)")
 
+## help: Get more info on make commands.
+help: Makefile
+	@echo " Choose a command run in "$(PROJECTNAME)":"
+	@sed -n 's/^##//p' $< | column -t -s ':' |  sed -e 's/^/ /'
 .PHONY: help
-help: Makefile ## Display all available make commands.
-	@echo " Choose a command to run in "$(PROJECTNAME)":"
-	@sed -n -e '/^.PHONY: /{N; s/^.PHONY: \(.*\)\n.*:.*## \(.*\)/\1:\2/; p}' $(MAKEFILE_LIST) | \
-	sort | \
-	awk 'BEGIN {FS = ":"; } {printf " \033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
-.PHONY: proto-gen-docker
-proto-gen: ## Generate protobuf files. Requires docker.
+## proto-gen: Generate protobuf files. Requires docker.
+proto-gen:
 	@echo "--> Generating Protobuf files"
 	$(DOCKER_BUF) generate
+.PHONY: proto-gen-docker
 
-.PHONY: proto-lint
-proto-lint: ## Lint protobuf files. Requires docker.
+## proto-lint: Lint protobuf files. Requires docker.
+proto-lint:
 	@echo "--> Linting Protobuf files"
 	@$(DOCKER_BUF) lint
+.PHONY: proto-lint
 
-.PHONY: lint
-lint: ## Lint source code. Requires golangci-lint.
+## lint: Lint Go files. Requires golangci-lint.
+lint:
 	@echo "--> Lint source code using golangci-lint"
 	@golangci-lint run
+.PHONY: lint
 
-.PHONY: test
-test: ## Run unit tests.
+## test: Run unit tests.
+test:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -12,3 +12,21 @@ proto-lint:
 	@echo "--> Linting Protobuf files"
 	@$(DOCKER_BUF) lint
 .PHONY: proto-lint
+
+## lint: Lint source code. Requires golangci-lint.
+lint:
+	@echo "--> Lint source code using golangci-lint"
+	@golangci-lint run
+.PHONY: lint
```
rootulp, line 20: [optional] we can add a section to the README that instructs developers to install golangci-lint. Something like https://github.com/celestiaorg/celestia-app/blob/b1592c314d4dca77c477e62ee1890970c6ee6e0c/README.md?plain=1#L119-L121

rootulp (dismissed):
Awesome, thanks for the contribution! We may consider also adding: `make help`. See [here](https://github.com/celestiaorg/celestia-app/blob/b1592c314d4dca77c477e62ee1890970c6ee6e0c/Makefile#L21-L25) but that could be a follow-up PR.

Makefile
00x-dx: @rootulp I have modified the same PR with readme as well as help command. Let me know your feedback. for the Makefile, i have followed the default format, and the commands would refer to .PHONY.

Makefile
```diff
@@ -1,32 +1,35 @@
...
-## proto-gen-docker: Generate protobuf files. Requires docker.
-proto-gen:
+.PHONY: help
+help: Makefile ## Display all available make commands
+	@echo " Choose a command run in "$(PROJECTNAME)":"
+	@sed -n -e '/^.PHONY: /{N; s/^.PHONY: \(.*\)\n.*:.*## \(.*\)/\1:\2/; p}' $(MAKEFILE_LIST) | \
+	sort | \
+	awk 'BEGIN {FS = ":"; } {printf " \033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+.PHONY: proto-gen-docker
```
rootulp, line 12:
```suggestion
.PHONY: proto-gen
```

Makefile
```diff
@@ -1,32 +1,35 @@
 DOCKER := $(shell which docker)
 DOCKER_BUF := $(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace bufbuild/buf:1.28.1
+PROJECTNAME=$(shell basename "$(PWD)")
 
-## proto-gen-docker: Generate protobuf files. Requires docker.
-proto-gen:
+.PHONY: help
+help: Makefile ## Display all available make commands
```
rootulp: This isn't working as expected

```
$ make help
 Choose a command to run in go-square:
sed: 1: "/^.PHONY: /{N; s/^.PHON ...": extra characters at the end of p command
```

rootulp: Pushed a fix...

```shell
$ make help
 Choose a command run in go-square:
  help         Get more info on make commands.
  proto-gen    Generate protobuf files. Requires docker.
  proto-lint   Lint protobuf files. Requires docker.
  lint         Lint Go files. Requires golangci-lint.
  test         Run unit tests.
  benchmark    Run tests in benchmark mode.
```

Makefile
rootulp: pushed a fix.

rootulp (approved):

Makefile
00x-dx: Thanks!

Makefile
rootulp: thanks, we appreciate it!

Makefile
00x-dx: That's interesting! May i know which env you are on ? I tested it on Pop_os though ..

rootulp: oh strange, maybe our `sed`s are incompatible? I'm on MacOS arm64. Unfortunately my sed doesn't tell me which version it is:

```
$ sed --version
sed: illegal option -- -
usage: sed script [-Ealnru] [-i extension] [file ...]
	sed [-Ealnu] [-i extension] [-e script] ... [-f script_file] ... [file ...]
```

Any chance you can test the latest version of this PR on your machine?

rootulp: Now that you mention it, I remember something may be weird with MacOS `sed` because I hit this: https://github.com/celestiaorg/celestia-app/blob/5476636002d4c4be029718230d5f8f6d321e3b6f/scripts/single-node.sh#L64-L67

00x-dx: Mine is 4.8. Recent change is working fine for me too.

rootulp: Okay awesome, thanks!

distractedm1nd (approved):

PR #42, opened 2024-02-29
shares: use more efficient make([]byte, N) not bytes.Repeat([]byte{0}, N)

odeke-em:
This change uses Go's memory model guarantees that the zero value memory shall be zeroed and hence to create a slice of 0 bytes aka padding, we can simply use make([]byte, N) which is much more efficient than bytes.Repeat([]byte{0}, N) and the benchmarks show this improvement:

```shell
$ benchstat before.txt after.txt
name               old time/op    new time/op    delta
PaddingVsRepeat-8     674ns ± 1%     538ns ± 0%  -20.22%  (p=0.000 n=9+9)

name               old alloc/op   new alloc/op   delta
PaddingVsRepeat-8    1.31kB ± 0%    0.83kB ± 0%  -36.59%  (p=0.000 n=10+10)

name               old allocs/op  new allocs/op  delta
PaddingVsRepeat-8      12.0 ± 0%      11.0 ± 0%   -8.33%  (p=0.000 n=10+10)
```


shares/utils.go
```diff
@@ -31,8 +31,7 @@ func zeroPadIfNecessary(share []byte, width int) (padded []byte, bytesOfPadding
 	}
 
 	missingBytes := width - oldLen
-	padByte := []byte{0}
-	padding := bytes.Repeat(padByte, missingBytes)
+	padding := make([]byte, missingBytes)
```
rootulp, line 34: LGTM because `Test_zeroPadIfNecessary` still passes

shares/padding.go
```diff
@@ -20,7 +19,7 @@ func NamespacePaddingShare(ns namespace.Namespace, shareVersion uint8) (Share, e
 	if err := b.WriteSequenceLen(0); err != nil {
 		return Share{}, err
 	}
-	padding := bytes.Repeat([]byte{0}, FirstSparseShareContentSize)
+	padding := make([]byte, FirstSparseShareContentSize)
```
rootulp, line 22: LGTM b/c `TestNamespacePaddingShare` still passes

rootulp (approved):
Thanks for the PR! I updated the PR title to conform to conventional commits to pass a CI check

liamsi (approved):

odeke-em:
Thank you for the review and merge!

PR #44, opened 2024-02-29
perf(merkle): optimize hashing by reusing sha256 in a pool

### LINKED ISSUE #43: merkle: innerHash dominates CPU/RAM consumption by sha256.Sum256(append(bytes...)) but it can be further sped up and allocations reduced as well by using a pool and reusing hashes
### Summary of Bug

I was auditing the code for both security vulnerabilities and performance improvements and after reading through the literature on HashFromByteSlicesIterative at https://github.com/celestiaorg/go-square/blob/c20eda971e90d33bd45fd44ca12ba9a9da7cf2d2/merkle/tree.go#L62 I saw that the prior speed ups had hit their limit; I also profiled the code and indeed confirmed that innerHash consumes lots of CPU and RAM too hence why the use of the recursive vs iterative hashing doesn't show significant differences; the profiles indeed confirmed that Amdahl's law does hold

Screenshot 2024-02-29 at 7 19 07 PM

## Analysis

If we examine the code for innerHash https://github.com/celestiaorg/go-square/blob/c20eda971e90d33bd45fd44ca12ba9a9da7cf2d2/merkle/hash.go#L23-L31

we can notice these interesting properties https://github.com/celestiaorg/go-square/blob/c20eda971e90d33bd45fd44ca12ba9a9da7cf2d2/merkle/hash.go#L29

that code uses a global call sha256.Sum256 whose code creates a new digest everytime and discards it per https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/crypto/sha256/sha256.go;l=253 
Screenshot 2024-02-29 at 7 21 33 PM

and right there is our opportunity!

## Remedy and optimization
We can reuse hashes and reset them each time, but even more, we can avoid the call to append each time but instead invoke `h.Write(byteslice)`

by this code
```go
var sha256Pool = &sync.Pool{New: func() any { return sha256.New() }}
        
func innerHashPooled(left, right []byte) []byte {
        h := sha256Pool.Get().(hash.Hash)
        defer sha256Pool.Put(h)
                        
        h.Reset()
        h.Write(innerPrefix)
        h.Write(left)
        h.Write(right)  
                        
        return h.Sum(nil)
}          
```

## Results
When we use that new routine, voila the results are remarkable
```shell
$ benchstat before.txt after.txt 
name                old time/op    new time/op    delta
HashAlternatives-8    79.9ms ± 3%    73.8ms ± 3%   -7.55%  (p=0.000 n=10+10)

name                old alloc/op   new alloc/op   delta
HashAlternatives-8    28.0MB ± 0%    13.6MB ± 0%  -51.42%  (p=0.000 n=9+10)

name                old allocs/op  new allocs/op  delta
HashAlternatives-8      500k ± 0%      300k ± 0%  -40.00%  (p=0.000 n=10+10)
```
Screenshot 2024-02-29 at 7 27 01 PM

of which we shaved off 1+ second comparably against leafHash :-)

/cc @elias-orijtech @rootulp @liamsi @musalbas 

### Version

4135f359b20ddd97b6cb74ecc831f70ffb73f8a5

### Steps to Reproduce

Please read the above!

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

odeke-em:
This change optimizes SHA256 hashing for leafHash & innerHash by pooling and reusing hashes, which massively reduces on RAM consumption and even CPU time (benchmarking on my noisy machine is skewed but the results can be verified individually) producing results below:

```shell
$ benchstat before.txt after.txt
name                          old time/op    new time/op    delta
HashAlternatives/recursive-8    77.6µs ± 2%    77.0µs ± 2%     ~     (p=0.165 n=10+10)
HashAlternatives/iterative-8    76.3µs ± 1%    76.2µs ± 3%     ~     (p=0.720 n=9+10)

name                          old alloc/op   new alloc/op   delta
HashAlternatives/recursive-8    25.4kB ± 0%     6.4kB ± 0%  -74.94%  (p=0.000 n=10+10)
HashAlternatives/iterative-8    28.1kB ± 0%     9.1kB ± 0%  -67.78%  (p=0.000 n=10+10)

name                          old allocs/op  new allocs/op  delta
HashAlternatives/recursive-8       497 ± 0%       199 ± 0%  -59.96%  (p=0.000 n=10+10)
HashAlternatives/iterative-8       498 ± 0%       200 ± 0%  -59.84%  (p=0.000 n=10+10)
```

Fixes #43

## Checklist

- [ ] New and updated code has appropriate documentation
- [ ] New and updated code has new and/or updated testing
- [ ] Required CI checks are passing
- [ ] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

### DIFF
--- a/merkle/hash.go
+++ b/merkle/hash.go
@@ -2,6 +2,8 @@ package merkle
 
 import (
 	"crypto/sha256"
+	shash "hash"
+	"sync"
 )
 
 // TODO: make these have a large predefined capacity
@@ -17,15 +19,26 @@ func emptyHash() []byte {
 
 // returns sha256(0x00 || leaf)
 func leafHash(leaf []byte) []byte {
-	return hash(append(leafPrefix, leaf...))
+	return hash(leafPrefix, leaf)
 }
 
 // returns sha256(0x01 || left || right)
-func innerHash(left []byte, right []byte) []byte {
-	return hash(append(innerPrefix, append(left, right...)...))
+func innerHash(left, right []byte) []byte {
+	return hash(innerPrefix, left, right)
 }
 
-func hash(bz []byte) []byte {
-	h := sha256.Sum256(bz)
-	return h[:]
+var sha256Pool = &sync.Pool{New: func() any { return sha256.New() }}
+
+func hash(slices ...[]byte) []byte {
+	h := sha256Pool.Get().(shash.Hash)
+	defer func() {
+		h.Reset()
+		sha256Pool.Put(h)
+	}()
+
+	for _, slice := range slices {
+		h.Write(slice)
+	}
+
+	return h.Sum(nil)
 }
--- a/merkle/tree_test.go
+++ b/merkle/tree_test.go
@@ -121,12 +121,14 @@ func BenchmarkHashAlternatives(b *testing.B) {
 
 	b.ResetTimer()
 	b.Run("recursive", func(b *testing.B) {
+		b.ReportAllocs()
 		for i := 0; i < b.N; i++ {
 			_ = HashFromByteSlices(items)
 		}
 	})
 
 	b.Run("iterative", func(b *testing.B) {
+		b.ReportAllocs()
 		for i := 0; i < b.N; i++ {
 			_ = HashFromByteSlicesIterative(items)
 		}


rootulp (commented):
These changes seem fine to me but it's difficult to approve because the modified functions don't have unit tests. They should've had unit tests prior to this PR but I expect they don't because the `merkle/` package is copy + pasted from the CometBFT repo. 

IMO we need to decide if it's okay for the `merkle` package in this repo to diverge from the upstream CometBFT implementation. cc: @cmwaters

odeke-em:
@rootulp how come the code was copied and pasted in here without the appropriate attributions/licenses or without just plain imports or submodules?

rootulp:
I think this is relevant:

https://github.com/celestiaorg/go-square/blob/4e84d80f3e55ac3093005c8089186d68ee62541d/merkle/README.md?plain=1#L6

I suppose we could also copy + paste [this license](https://github.com/cometbft/cometbft/blob/main/LICENSE) into that directory.

cmwaters:
Closing this because for now we've decided to keep the merkle library in our cometbft fork. Thanks for opening the PR in any case. Sorry that we didn't merge it. We may look to apply this to our cometbft fork in the future

Issue #47, opened 2024-03-05
bug: worst case share indexes uses max square size


rootulp:
## Context

https://github.com/celestiaorg/go-square/blob/a84173978ed29049469dc5e84b4a58cfece98401/square/builder.go#L414

## Problem

We want worstCaseShareIndexes to always use the upper bound for max square size (128) so that celestia-node doesn't have to be stateful when extending the ODS.

If `worstCaseShareIndexes` uses the max effective square size (64) then celestia-node would have to poll celestia-app for the gov max square size every block.

Also note that celestia-app v1.x used a worstCaseShareIndexes of `128` so it was a consensus breaking change to drop down to `64`.

## Proposal

Modify the second argument to this function to use the upper bound of `128`

rootulp:
FWIW I think this issue has existed since go-square was extracted from celestia-app. See https://github.com/celestiaorg/go-square/commit/68863fc876b175343fe68e21467c70652e629280#diff-f37b1db98350d69502a30f6f3dd88ae2f6844b2bb6e0c46f89d3b6a237788b3cR414

rootulp:
I just thought of a way to fix this bug without introducing an API breaking change, will publish an alternative PR soon.

PR #48, opened 2024-03-05
fix!: use squareSizeUpperBound for worstCaseShareIndexes

rootulp:
Closes https://github.com/celestiaorg/go-square/issues/47

This restores the behavior from celestia-app v1.x where square size upper bound is used to calculate worstCaseShareIndexes. 

## Testing

In celestia-app, TestMaxBlockSize flake is no longer observed

```
 go test -run TestIntegrationTestSuite/TestMaxBlockSize -count 5 -v
```

### DIFF
--- a/square/builder.go
+++ b/square/builder.go
@@ -16,6 +16,8 @@ import (
 type Builder struct {
 	// maxSquareSize is the maximum number of rows (or columns) in the original data square
 	maxSquareSize int
+	// squareSize imposes an upper bound on the max effective square size.
+	squareSizeUpperBound int
 	// currentSize is an overestimate for the number of shares used by this builder.
 	currentSize int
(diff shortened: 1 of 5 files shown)


rootulp:
Note to reviewers: we could add a unit test to this PR that includes a `bad_block.json` from a failing TestMaxBlockSize test case

evan-forbes (approved):

rootulp:
We could merge this as-is but I want @cmwaters 's review. I'm also nervous about regressions in this repo because this fix was added without a test that would fail if we ever revert this behavior 😞 

I'll work on adding a test now.

rootulp:
Closing in favor of https://github.com/celestiaorg/go-square/pull/49

PR #49, opened 2024-03-09
fix: worst case share indexes

rootulp:
Closes https://github.com/celestiaorg/go-square/issues/47 via a non-API breaking change. This approach has the downside that it duplicates a constant from celestia-app and we'll need a way to plumb the constant through or refactor the implementation here if/when we actually modify that versioned constant in celestia-app.


square/builder_test.go
```diff
@@ -341,3 +343,33 @@ func generateBlobTxsWithNamespaces(namespaces []namespace.Namespace, blobSizes [
...
+	require.NoError(t, err)
+
+	builder, err := square.NewBuilder(defaultMaxSquareSize, defaultSubtreeRootThreshold, bigBlock.Txs...)
+	require.NoError(t, err)
+	assert.Len(t, builder.Blobs, 84)
+	assert.Len(t, builder.Pfbs, 25)
+
+	index, err := builder.FindBlobStartingIndex(0, 0)
+	require.NoError(t, err)
+	assert.Equal(t, 2234, index)
```
rootulp, line 371: this line fails if the implementation uses `64 * 64` when calculating the worst case share indexes because that would result in slightly fewer bytes being accounted for in the compact shares at the beginning of the square which would move the blob share index back a little bit 

```
--- FAIL: TestBigBlock (0.04s)
    builder_test.go:371: 
        	Error Trace:	/Users/rootulp/git/rootulp/celestiaorg/go-square/square/builder_test.go:371
        	Error:      	Not equal: 
        	            	expected: 2234
        	            	actual  : 2232
        	Test:       	TestBigBlock
FAIL
```

cmwaters (approved):
Yeah this makes sense as a good non breaking solution for v1

EDIT: Actually I think this theoretically breaks the byte format of the square and so is thus still considered breaking

evan-forbes (approved):

PR #63, opened 2024-05-29
perf(namespace): speed up Namespace comparisons

cristaloleg:
## Overview

While reading and checking structures from https://celestiaorg.github.io/celestia-app/specs/data_structures.html I noticed that we compare namespaces not optimal enough. Every equal or compare-like operation allocates new slice with data that can be already compared directly. 

```
% go-perftuner bstat a.txt b.txt
args: [a.txt b.txt]name        old time/op    new time/op    delta
Equal-10      71.9ns ± 0%     3.9ns ± 0%   -94.53%  (p=0.008 n=5+5)
Compare-10    72.0ns ± 1%     4.4ns ± 5%   -93.83%  (p=0.002 n=6+6)

name        old alloc/op   new alloc/op   delta
Equal-10       64.0B ± 0%      0.0B       -100.00%  (p=0.026 n=5+6)
Compare-10     64.0B ± 0%      0.0B       -100.00%  (p=0.002 n=6+6)

name        old allocs/op  new allocs/op  delta
Equal-10        2.00 ± 0%      0.00       -100.00%  (p=0.026 n=5+6)
Compare-10      2.00 ± 0%      0.00       -100.00%  (p=0.002 n=6+6)
```

### DIFF
--- a/namespace/namespace.go
+++ b/namespace/namespace.go
@@ -119,23 +119,23 @@ func (n Namespace) IsSecondaryReserved() bool {
 }
 
 func (n Namespace) IsParityShares() bool {
-	return bytes.Equal(n.Bytes(), ParitySharesNamespace.Bytes())
+	return n.Equals(ParitySharesNamespace)
 }
 
 func (n Namespace) IsTailPadding() bool {
-	return bytes.Equal(n.Bytes(), TailPaddingNamespace.Bytes())
+	return n.Equals(TailPaddingNamespace)
 }
 
 func (n Namespace) IsPrimaryReservedPadding() bool {
-	return bytes.Equal(n.Bytes(), PrimaryReservedPaddingNamespace.Bytes())
+	return n.Equals(PrimaryReservedPaddingNamespace)
 }
 
 func (n Namespace) IsTx() bool {
-	return bytes.Equal(n.Bytes(), TxNamespace.Bytes())
+	return n.Equals(TxNamespace)
 }
 
 func (n Namespace) IsPayForBlob() bool {
-	return bytes.Equal(n.Bytes(), PayForBlobNamespace.Bytes())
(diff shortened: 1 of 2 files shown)


namespace/namespace.go
```diff
@@ -147,23 +147,34 @@ func (n Namespace) Repeat(times int) []Namespace {
...
-	return bytes.Compare(n.Bytes(), n2.Bytes()) > -1
+	return n.Compare(n2) > -1
+}
+
+func (n Namespace) Compare(n2 Namespace) int {
+	switch {
+	case n.Version == n2.Version:
+		return bytes.Compare(n.ID, n2.ID)
+	case n.Version < n2.Version:
+		return -1
```
cristaloleg, line 174: Later today I will check that -1 and 1 are not mixed here 🙈

rootulp: One way to test is to add a unit test for `Compare` and ensure that the output from before this change matches the output from after this PR because AFAIU this is strictly a performance optimization but behavior should remain the same.

rootulp (commented):
Great work, thanks for the PR! Before approving, I think unit tests would really help our confidence that behavior remains the same.

[blocking] unit test for `Compare` would really help confidence.
[optional] unit tests for `IsLessThan`, `IsLessOrEqualThan`, `IsGreaterThan`, `IsGreaterOrEqualThan`

cristaloleg:
Thanks everyone, will address comments tomorrow 🫡

namespace/namespace.go
cristaloleg: Done, `Test_compareMethods` covers all code branches now.

cmwaters (approved):
lgtm. Thanks for also adding the tests

rootulp (approved):

PR #64, opened 2024-05-30
fix(blob): use proper share.MaxShareVersion const

cristaloleg:
## Overview

We were comparing against incorrect const (`math.MaxUint8`) but error message refers to the proper (`share.MaxShareVersion`).

Sadly we cannot import `shares` package in `blob` due to cyclic dependency, adding comment for the future. Also, removed useless `fmt` alias for `fmt` package.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -23,6 +23,9 @@ const ProtoBlobTxTypeID = "BLOB"
 // decoding binaries that are not actually IndexWrappers.
 const ProtoIndexWrapperTypeID = "INDX"
 
+// MaxShareVersion is the maximum value a share version can be. See: [shares.MaxShareVersion].
+const MaxShareVersion = 127
+
 // New creates a new coretypes.Blob from the provided data after performing
 // basic stateless checks over it.
 func New(ns namespace.Namespace, blob []byte, shareVersion uint8) *Blob {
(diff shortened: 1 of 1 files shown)


cmwaters (approved):
Good catch!

Is there a cyclic dependency when importing the constant from the shares package or why can't we just use the imported value like we do with namespace

cristaloleg:
I mentioned cyclic import in PR description.

Wondertan:
> https://github.com/celestiaorg/go-square/pull/1#issuecomment-1857914345

I thought I mentioned it in that comment, but I likely forgot. Another good reason to keep core-core types under the same pkg is that we wouldn't need to deal with cycles.

blob/blob.go
```diff
@@ -51,7 +50,7 @@ func (b *Blob) Validate() error {
 	if len(b.NamespaceId) != namespace.NamespaceIDSize {
 		return fmt.Errorf("namespace id must be %d bytes", namespace.NamespaceIDSize)
 	}
-	if b.ShareVersion > math.MaxUint8 {
+	if b.ShareVersion > 127 { // see: share.MaxShareVersion
```
rootulp, line 53: [optional] should we re-define the constant in this package perhaps at the top of this file? Something like:

```go

// MaxShareVersion is the maximum value a share version can be.
// Note this is copied from share.MaxShareVersion to avoid an import cycle.
MaxShareVersion = 127
```

then here

```suggestion
	if b.ShareVersion > MaxShareVersion {
```

walldiss (approved):

rootulp (approved):
Good catch! Thanks for fixing. This could use a unit test to prevent regressions but not blocking on it.

blob/blob.go
cristaloleg: Good point, done.

cmwaters (approved):

ninabarbakadze (approved):

Issue #66, opened 2024-06-10
Canonicalise Namespace and Share types


cmwaters:
### Summary

We have different constructions of the Share and Namespace types across celestia repos. We should converge on types and have them standardised across repos.

https://github.com/celestiaorg/celestia-node/blob/762bd93b1b6242bdd5660172d1b32fcade144a8c/share/namespace.go#L34

https://github.com/celestiaorg/go-square/blob/fd78256df076539ef5e13cd271276f058a167d8c/namespace/namespace.go#L8-L11

https://github.com/celestiaorg/celestia-node/blob/762bd93b1b6242bdd5660172d1b32fcade144a8c/share/share.go#L29

https://github.com/celestiaorg/go-square/blob/fd78256df076539ef5e13cd271276f058a167d8c/shares/shares.go#L12-L14

### Proposal

Agreed upon types:

```go
type Namespace struct {
    data []byte
}

type Share struct {
    data []byte
}
```

We have decided to encapsulate the types as this completely prevents users from misusing them (i.e. incorrect length).

Encapsulation also allows us to make changes in the future that don't break users such as caching certain bytes in the struct i.e. namespace in the share struct.

The other option considered was to use `type Namespace []byte` 

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #67, opened 2024-06-11
chore: remove go.work and go.work.sum

cmwaters:

PR #68, opened 2024-06-11
feat!: refactor Namespace type

cmwaters:
Addresses: https://github.com/celestiaorg/go-square/issues/66


rootulp (approved):

blob/blob.go
```diff
@@ -30,19 +30,16 @@ const MaxShareVersion = 127
 // basic stateless checks over it.
 func New(ns namespace.Namespace, blob []byte, shareVersion uint8) *Blob {
 	return &Blob{
-		NamespaceId:      ns.ID,
+		NamespaceId:      ns.ID(),
 		Data:             blob,
 		ShareVersion:     uint32(shareVersion),
-		NamespaceVersion: uint32(ns.Version),
+		NamespaceVersion: uint32(ns.Version()),
```
walldiss, line 36: Do you think it make sense to rework blob in this PR too, so it contains new Namespace, rather than NamespaceID and NamespaceVersion as separate fields?

namespace/consts.go
```diff
@@ -70,14 +70,12 @@ var (
...
-		ID:      append(bytes.Repeat([]byte{0x00}, NamespaceIDSize-1), lastByte),
+		data: append([]byte{NamespaceVersionZero}, append(bytes.Repeat([]byte{0x00}, NamespaceIDSize-1), lastByte)...),
 	}
 }
 
 func secondaryReservedNamespace(lastByte byte) Namespace {
 	return Namespace{
-		Version: NamespaceVersionMax,
-		ID:      append(bytes.Repeat([]byte{0xFF}, NamespaceIDSize-1), lastByte),
+		data: append([]byte{NamespaceVersionMax}, append(bytes.Repeat([]byte{0xFF}, NamespaceIDSize-1), lastByte)...),
```
walldiss, line 79: There are different approaches to do the same within this PR. In another place it is done by:
```
data := []byte{version}
data = append(data, id...)
```
I personally find later to read easier.

It looks like local constructor without inner validation will be handy. It will remove the need to do append tricks every time and unify the construction:
`func newNamespace(v version, id []bytes)`

blob/blob.go
```diff
@@ -105,7 +102,9 @@ func MarshalBlobTx(tx []byte, blobs ...*Blob) ([]byte, error) {
 // Sort sorts the blobs by their namespace.
 func Sort(blobs []*Blob) {
 	sort.SliceStable(blobs, func(i, j int) bool {
-		return bytes.Compare(blobs[i].Namespace().Bytes(), blobs[j].Namespace().Bytes()) < 0
+		ns1 := append([]byte{byte(blobs[i].NamespaceVersion)}, blobs[i].NamespaceId...)
+		ns2 := append([]byte{byte(blobs[j].NamespaceVersion)}, blobs[j].NamespaceId...)
+		return bytes.Compare(ns1, ns2) < 0
```
cristaloleg, line 107: Why not add method `Compare(other *Blob)` to `Blob` type instead?

blob/blob.go
cmwaters: I thought about this and am open to discussion. This would break the state machine which I'm hesitant to do without sufficient reason

namespace/consts.go
cmwaters: Makes sense. I also have a few ideas of constructing the namespace from just the bytes instead of needing to pass them separately (similar to `From`)

blob/blob.go
cmwaters: Yeah I can do that

blob/blob.go
walldiss: Digging more into it I've found that Blob struct is actually generated from proto. So proper way to make it would be to create local struct, that has fields and methods for it, while making proto as a wire representation only and putting it into separate pkg. Lets leave it as is and create another issue with canonicalise Blob.

blob/blob.go
```diff
@@ -30,19 +30,23 @@ const MaxShareVersion = 127
...
 
 // Namespace returns the namespace of the blob
-func (b *Blob) Namespace() namespace.Namespace {
-	return namespace.Namespace{
-		Version: uint8(b.NamespaceVersion),
-		ID:      b.NamespaceId,
-	}
+func (b *Blob) Namespace() (namespace.Namespace, error) {
+	return namespace.NewFromBytes(b.RawNamespace())
+}
```
walldiss, line 43: It seems we expect Blob to have valid namespace at that point? If so, we can avoid using exported constractor in favour of local(without validion) constractor. I will remove the need to return error here and:
- Method will be be much easier to use without `if err != nil ...`
- Won't need `(b *Blob) RawNamespace()` anymore

cristaloleg (approved):
SGTM

namespace/namespace.go
```diff
@@ -39,6 +42,27 @@ func MustNew(version uint8, id []byte) Namespace {
...
+	if len(bytes) != NamespaceSize {
+		return Namespace{}, fmt.Errorf("invalid namespace length: %v must be %v", len(bytes), NamespaceSize)
+	}
+
+	err := validateVersionSupported(bytes[0])
+	if err != nil {
+		return Namespace{}, err
+	}
+
+	err = validateID(bytes[0], bytes[1:])
```
walldiss, line 56: Looks like validateID taking 2 params is an artifact from the previous implementation, perhaps validateID can take just full `bytes` now instead of subsclices.

namespace/namespace.go
```diff
@@ -23,10 +22,14 @@ func New(version uint8, id []byte) (Namespace, error) {
...
+}
+
+func newNamespace(version uint8, id []byte) Namespace {
+	data := []byte{version}
 	return Namespace{
-		Version: version,
-		ID:      id,
-	}, nil
+		data: append(data, id...),
+	}
```
Wondertan, line 32: Allocation police 🚓 
This allocates twice, but we could do it in a single allocation. One of the way to do that is similar to https://github.com/celestiaorg/celestia-node/blob/540192259c144ccbd24e45e34616a41389232a51/share/namespace.go#L46-L50

blob/blob.go
```diff
@@ -30,19 +30,23 @@ const MaxShareVersion = 127
...
+
+// Namespace returns the namespace of the blob
+func (b *Blob) RawNamespace() []byte {
+	namespace := []byte{byte(b.NamespaceVersion)}
+	namespace = append(namespace, b.NamespaceId...)
+	return namespace
 }
 
 // Validate runs a stateless validity check on the form of the struct.
```
Wondertan, line 52: Besides Validate, the node needs two more validate funcs. https://github.com/celestiaorg/celestia-node/blob/540192259c144ccbd24e45e34616a41389232a51/share/namespace.go#L114-L129.

We need this so we can restrict to only blobs(excluding reserved) and to only real data(excluding parity)

blob/blob.go
cmwaters: Yeah I was thinking the same that way you only verify once upon converting to the local type

blob/blob.go
cmwaters: Sure, in a follow up PR

namespace/namespace.go
cmwaters: In `New` we still pass two arguments (ID and Version). `validateID` matches that function signature

blob/blob.go
cmwaters: Let me extract this out as a separate issue. I imagine there's some overlap with celestia-app and the validation there

namespace/namespace.go
```diff
@@ -70,18 +96,24 @@ func MustNewV0(subID []byte) Namespace {
...
 
 // Bytes returns this namespace as a byte slice.
 func (n Namespace) Bytes() []byte {
-	return append([]byte{n.Version}, n.ID...)
+	return n.data
+}
+
+// Version return this namespace's version
+func (n Namespace) Version() uint8 {
+	return n.data[0]
```
Wondertan, line 111: I think having zero here is fine, but we could also avoid having a magic number here and use the constant like we do [here](https://github.com/celestiaorg/celestia-node/blob/540192259c144ccbd24e45e34616a41389232a51/share/namespace.go#L61)

namespace/namespace.go
```diff
@@ -7,8 +7,7 @@ import (
 )
 
 type Namespace struct {
-	Version uint8
-	ID      []byte
+	data []byte
 }
 
 // New returns a new namespace with the provided version and id.
```
Wondertan, line 13: I recommend having a versioned constructor. This is helpful for devs to construct the blob with a namespace. It's usually unclear where to get the version constant and which one to use, while a constructor for every version(along general one) simplifies that. Considering that new versions aren't coming out often, adding a new constructor for each would have add much overhead.

Wondertan: Besides, each constructor may address the quirks of a namespace. For example, V0 requires first N zero bytes for the ID, and expecting the user to know that to construct the correct ID is ridiculous. The specialized constructor can expect exactly the needed amount of bytes that the user controls.

cmwaters: There is already `NewV0`

Wondertan: Missed that.

Wondertan: Can we make sure NewV0 allocates once as well? Currently, it does that twice in itself and in the `newNamespace`?

blob/blob.go
```diff
@@ -39,7 +39,15 @@ func New(ns namespace.Namespace, blob []byte, shareVersion uint8) *Blob {
 
 // Namespace returns the namespace of the blob
 func (b *Blob) Namespace() (namespace.Namespace, error) {
-	return namespace.New(uint8(b.NamespaceVersion), b.NamespaceId)
+	return namespace.NewFromBytes(b.RawNamespace())
+}
+
+// Namespace returns the namespace of the blob
```
rootulp, line 45:
```suggestion
// RawNamespace returns the raw namespace of this blob.
```

rootulp (dismissed):

walldiss (dismissed):

namespace/namespace.go
Wondertan: The main reason is that this is worse then the node's status quo and migrating over it is meh

namespace/namespace.go
```diff
@@ -39,6 +44,27 @@ func MustNew(version uint8, id []byte) Namespace {
...
 
+// NewFromBytes returns a new namespace from the provided byte slice.
+func NewFromBytes(bytes []byte) (Namespace, error) {
+	if len(bytes) != NamespaceSize {
+		return Namespace{}, fmt.Errorf("invalid namespace length: %v must be %v", len(bytes), NamespaceSize)
+	}
+
+	err := validateVersionSupported(bytes[0])
+	if err != nil {
+		return Namespace{}, err
```
cristaloleg, line 55: Should we wrap error with additional info what was wrong? Something like:

```
return Namespace{}, fmt.Errorf("invalid version: %v", bytes[0])
```

Same for the `validateID` below.

blob/blob.go
```diff
@@ -30,19 +30,24 @@ const MaxShareVersion = 127
...
-		ID:      b.NamespaceId,
-	}
+func (b *Blob) Namespace() (namespace.Namespace, error) {
+	return namespace.NewFromBytes(b.RawNamespace())
+}
+
+// Namespace returns the namespace of the blob
+func (b *Blob) RawNamespace() []byte {
+	namespace := make([]byte, namespace.NamespaceSize)
+	namespace[0] = uint8(b.NamespaceVersion)
```
cristaloleg, line 48: Just for the consistency: `byte` instead of `uint8`.

namespace/namespace.go
cmwaters: Have a look now

namespace/namespace.go
cmwaters: The errors seem descriptive enough themselves

namespace/namespace.go
```diff
@@ -70,18 +90,24 @@ func MustNewV0(subID []byte) Namespace {
...
 
 // Bytes returns this namespace as a byte slice.
 func (n Namespace) Bytes() []byte {
-	return append([]byte{n.Version}, n.ID...)
+	return n.data
+}
+
+// Version return this namespace's version
+func (n Namespace) Version() uint8 {
+	return n.data[NamespaceVersionSize-1]
```
renaynay, line 105: why not NamespaceVersionIndex

blob/blob.go
```diff
@@ -28,35 +28,40 @@ const MaxShareVersion = 127
 
 // New creates a new coretypes.Blob from the provided data after performing
 // basic stateless checks over it.
-func New(ns namespace.Namespace, blob []byte, shareVersion uint8) *Blob {
+func New(ns ns.Namespace, blob []byte, shareVersion uint8) *Blob {
 	return &Blob{
-		NamespaceId:      ns.ID,
+		NamespaceId:      ns.ID(),
```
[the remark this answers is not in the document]
cmwaters: This is the proto generated struct and I don't intend to break that but I will keep your suggestion for when we create a go-native `Blob` implementation

namespace/consts.go
```diff
@@ -69,15 +73,9 @@ var (
 )
 
 func primaryReservedNamespace(lastByte byte) Namespace {
-	return Namespace{
-		Version: NamespaceVersionZero,
-		ID:      append(bytes.Repeat([]byte{0x00}, NamespaceIDSize-1), lastByte),
-	}
+	return newNamespace(NamespaceVersionZero, append(bytes.Repeat([]byte{0x00}, NamespaceIDSize-1), lastByte))
```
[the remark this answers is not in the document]
cmwaters: If we were to go back we would probably start user namespaces at v1 and have `NamespaceVersionMin` as 0 for the primary reserved namespaces. For now v0 is both used by external users and by us. I think it would be more confusing to call it version min

namespace/namespace.go
```diff
@@ -23,10 +22,16 @@ func New(version uint8, id []byte) (Namespace, error) {
 		return Namespace{}, err
 	}
 
+	return newNamespace(version, id), nil
+}
+
+func newNamespace(version uint8, id []byte) Namespace {
+	data := make([]byte, 1+len(id))
```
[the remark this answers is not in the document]
cmwaters: Yeah I missed these parts when going through it again

namespace/namespace.go
cmwaters: It was updated in two steps. I just missed this. Thanks for pointing it out

.gitignore
```diff
@@ -0,0 +1,2 @@
+go.work
+go.work.sum
```
Wondertan, line 2: Do we still need this, considering https://github.com/celestiaorg/go-square/pull/67 was merged?

blob/blob.go
```diff
@@ -65,6 +70,10 @@ func (b *Blob) Validate() error {
 	return nil
 }
 
+func (b *Blob) Compare(other *Blob) int {
+	return bytes.Compare(b.RawNamespace(), other.RawNamespace())
+}
```
Wondertan, line 75: Usually, such methods are called Equals with the bool return value, but as long as this is not PR targeting Blob we can omit that and this method altogether

Wondertan (dismissed):

.gitignore
cmwaters: I think so, else it's going to see this as a diff between the users branch and main and will want to check it in

blob/blob.go
cmwaters: Ref: https://github.com/celestiaorg/go-square/pull/68#discussion_r1635996282

namespace/consts.go
```diff
@@ -9,6 +9,10 @@ const (
 	// NamespaceVersionSize is the size of a namespace version in bytes.
 	NamespaceVersionSize = 1
 
+	// VersionIndex is the index of the version in the namespace. This should
+	// always be the first byte
+	VersionIndex = NamespaceVersionSize - 1 // 0
```
rootulp, line 14: nit: I don't think we need to define this constant in terms of `NamespaceVersionSize` because if we updated `NamespaceVersionSize` to 2 then this constant wouldn't make sense any more so having it auto update to `2 -1 = 1` isn't necessary or desirable. In other words, I think this could be:

```suggestion
	VersionIndex = 0
```

rootulp (approved):

rootulp (approved):

blob/blob.go
Wondertan: Thats improvement over that comment

Wondertan (approved):

blob/blob.go
cmwaters: I actually thinking comparing is more important because it's more common to want to order these things then see if the namespaces are equal

cmwaters: Of course nothing wrong with having both

namespace/consts.go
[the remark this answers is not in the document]
renaynay: Unfortunate but ok

PR #71, opened 2024-06-13
ci: utilize go version from go.mod

ramin:
## Overview

drive by! was glancing at the namespace PR and saw we could remove the hardcoded `GO_VERSION` ENV var for loading directly from the go.mod, figured i'd do my part 🫡


cristaloleg (approved):
Nice! Didn't know `./go.mod` syntax.

cmwaters (approved):
Ah I didn't realise you could do that, thanks!

PR #72, opened 2024-06-14
feat: add signer to blob proto

cmwaters:
First step is to simply add the signer to the protos. Then I will create the go native blob version and the constructors for v0 and v1 share versions


evan-forbes (approved):

blob/blob.proto
```diff
@@ -12,6 +12,7 @@ message Blob {
   bytes data = 2;
   uint32 share_version = 3;
   uint32 namespace_version = 4;
+  string signer = 5;
```
Wondertan, line 15: Why string?

rootulp: [question]

1. Should this be a `types.AccAddress` or `[]byte` instead of a string? Context: a sig verifiable transaction is of type `[]types.AccAddress`:

```go
// SigVerifiableTx defines a transaction interface for all signature verification
// handlers.
type SigVerifiableTx interface {
	types.Tx
	GetSigners() []types.AccAddress
	GetPubKeys() ([]cryptotypes.PubKey, error) // If signer already has pubkey in context, this list will have nil in its place
	GetSignaturesV2() ([]signing.SignatureV2, error)
}

// Tx defines a transaction interface that supports all standard message, signature
// fee, memo, tips, and auxiliary interfaces.
type Tx interface {
	SigVerifiableTx

	types.TxWithMemo
	types.FeeTx
	tx.TipTx
	types.TxWithTimeoutHeight
}
```

2. [optional] if we do use a string, should we describe what format the string should be in? In other words something like this:

```suggestion
  // Signer is the Bech32 human readable address of the tx signer that submitted this blob. 
  // Example: "celestia1kj39jkzqlr073t42am9d8pd40tgudc3e2kj9yf"
  string signer = 5;
```

cmwaters: > Why string?

This is common for SDK messages like `MsgSend`

cmwaters: > Should this be a types.AccAddress or []byte instead of a string? Context: a sig verifiable transaction is of type []types.AccAddress:

This would involve importing the sdk, which I think we wanted to avoid iirc. Alternatively, if people want, we can user `bytes` for the signer

> [optional] if we do use a string, should we describe what format the string should be in? In other words something like this:

Yes I'm fine with that

rootulp: > This is common for SDK messages like `MsgSend`

Good point. They don't describe in Go docs what format the addresses should be in:

```go
// MsgSend represents a message to send coins from one account to another.
type MsgSend struct {
	FromAddress string                                   `protobuf:"bytes,1,opt,name=from_address,json=fromAddress,proto3" json:"from_address,omitempty"`
	ToAddress   string                                   `protobuf:"bytes,2,opt,name=to_address,json=toAddress,proto3" json:"to_address,omitempty"`
	Amount      github_com_cosmos_cosmos_sdk_types.Coins `protobuf:"bytes,3,rep,name=amount,proto3,castrepeated=github.com/cosmos/cosmos-sdk/types.Coins" json:"amount"`
}
```

but I'm pretty confident it's the Bech32 human readable address because [Mintscan](https://www.mintscan.io/celestia/tx/A802E82D6F64F026C53C82A3D36A86929EA0F8CD52AA13064FF163BF8272AFE1?height=1691974&sector=json):

Screenshot 2024-06-17 at 10 33 09 AM

rootulp (approved):
Out of curiosity, do you think it will be easier to add the signer field to the existing `Blob` type (a.k.a this PR) or create an entirely new type like `BlobV2` with the signer field + all the fields in the current `Blob`.

blob/blob.proto
Wondertan: I'm not sure go-square has to follow cosmos semantics, but I hope we won't need to pull tx types in here.

Wondertan: Do we know if there is a reason why they do that? Intuitively, I would put that as bytes because I don't see where I would use an address string in a human-readable use case we have the blob type for.

cmwaters: bytes is also a lot more compact (half the size), so perhaps I'll shift to bytes then

cmwaters:
> Out of curiosity, do you think it will be easier to add the signer field to the existing `Blob` type (a.k.a this PR) or create an entirely new type like `BlobV2` with the signer field + all the fields in the current `Blob`.

I think they would be much the same. Proto was designed to be used this way (i.e. creating new fields in a backwards compatible manner)

Wondertan (approved):

cristaloleg (approved):
SGTM

evan-forbes (approved):

Issue #70, opened 2024-06-12
Add a local blob type


cmwaters:
It seems we expect Blob to have valid namespace at that point? If so, we can avoid using exported constractor in favour of local(without validion) constractor. I will remove the need to return error here and:
- Method will be be much easier to use without `if err != nil ...`
- Won't need `(b *Blob) RawNamespace()` anymore

_Originally posted by @walldiss in https://github.com/celestiaorg/go-square/pull/68#discussion_r1636267713_

PR #74, opened 2024-06-14
feat!: create native blob struct

cmwaters:
This builds on https://github.com/celestiaorg/go-square/pull/72

Closes: https://github.com/celestiaorg/go-square/issues/70

Note: this does not touch any of the share construction yet


cmwaters:
I haven't added a maximum size check to the length of the signer. `sdk.AccAddress` is usually 20 bytes (the first 20 of a users public key), but in the SDK, I see a check of up to 255 bytes. Should I add one in here. Note that in celestia-app, we will have a specific check to see if it's a valid sdk.AccAddress and matches the signer of the PFB

blob/blob.go
```diff
@@ -26,53 +25,83 @@ const ProtoIndexWrapperTypeID = "INDX"
 // MaxShareVersion is the maximum value a share version can be. See: [shares.MaxShareVersion].
 const MaxShareVersion = 127
 
+// Blob (stands for binary large object) is a core type that represents data
+// to be submitted to the Celestia network alongside an accompanying namespace
+// and optional signer (for proving the author of the blob)
```
rootulp, line 30: uber nit I think we should be careful to not conflate author and signer. A user can author a blob and hand it to a different entity (the blob signer) who signs the PFB tx and submits it on-chain. In other words, the _signer_ can be different from the _author_.

```suggestion
// and optional signer (for proving the signer of the blob)
```

shares/split_sparse_shares.go
```diff
@@ -22,22 +22,15 @@ func NewSparseShareSplitter() *SparseShareSplitter {
...
 
-	rawData := blob.Data
-	blobNamespace, err := blob.Namespace()
-	if err != nil {
-		return err
-	}
+	rawData := blob.Data()
+	blobNamespace := blob.Namespace()
 
 	// First share (note by validating the blob we can safely cast the share version to uint8)
```
rootulp, line 32: comment no longer seems necessary b/c there isn't an explicit `uint8()` cast

```suggestion
```

rootulp (approved):
LGTM, great work

blob/blob.go
cmwaters: As far as Celestia is concerned the signer is always the author but I agree that we should try keep the nomenclature the same throughout (i.e. we use `Signer` in the blob itself)

rootulp (approved):

square/builder.go
```diff
@@ -96,7 +96,13 @@ func (b *Builder) AppendBlobTx(blobTx *blob.BlobTx) bool {
...
 	blobElements := make([]*Element, len(blobTx.Blobs))
 	maxBlobShareCount := 0
-	for idx, blob := range blobTx.Blobs {
+	for idx, protoBlob := range blobTx.Blobs {
+		blob, err := blob.NewFromProto(protoBlob)
+		if err != nil {
+			// TODO: we should look at having a go native BlobTx type
+			// that we have already verified instead of doing it twice here
+			panic(fmt.Sprintf("invalid blob %d: %v", idx, err))
+		}
```
evan-forbes, line 105: Turning this into an issue

square/builder.go
```diff
@@ -96,7 +96,13 @@ func (b *Builder) AppendBlobTx(blobTx *blob.BlobTx) bool {
 	// create a new blob element for each blob and track the worst-case share count
 	blobElements := make([]*Element, len(blobTx.Blobs))
 	maxBlobShareCount := 0
-	for idx, blob := range blobTx.Blobs {
+	for idx, protoBlob := range blobTx.Blobs {
+		blob, err := blob.NewFromProto(protoBlob)
+		if err != nil {
+			// TODO: we should look at having a go native BlobTx type
+			// that we have already verified instead of doing it twice here
+			panic(fmt.Sprintf("invalid blob %d: %v", idx, err))
```
evan-forbes, line 104: [not blocking nit]

can we add add a comment to this issue to indicate that it requires that all blobs passed to it are valid? 

either that or add a new method / break this api that allows for error handling. my personal preference is that we panic as high in the stack as possible as that makes the problem that much clearer.

evan-forbes (approved):

square/builder.go
```diff
@@ -128,8 +134,8 @@ func (b *Builder) Export() (Square, error) {
 	// of blobs within a namespace because b.Blobs are already ordered by tx
 	// priority.
 	sort.SliceStable(b.Blobs, func(i, j int) bool {
-		ns1 := append([]byte{byte(b.Blobs[i].Blob.NamespaceVersion)}, b.Blobs[i].Blob.NamespaceId...)
-		ns2 := append([]byte{byte(b.Blobs[j].Blob.NamespaceVersion)}, b.Blobs[j].Blob.NamespaceId...)
+		ns1 := b.Blobs[i].Blob.Namespace().Bytes()
+		ns2 := b.Blobs[j].Blob.Namespace().Bytes()
 		return bytes.Compare(ns1, ns2) < 0
```
cristaloleg, line 139: Looks like this can be a `ns1.Compare(ns2) < 0` (https://github.com/celestiaorg/go-square/blob/main/namespace/namespace.go#L195)

shares/parse_sparse_shares.go
```diff
@@ -56,27 +59,29 @@ func parseSparseShares(shares []Share, supportedShareVersions []uint8) (blobs []
...
+				ns:           ns,
+				shareVersion: version,
+				data:         data,
+				sequenceLen:  sequenceLen,
 			})
 		} else { // continuation share
 			if len(sequences) == 0 {
 				return nil, fmt.Errorf("continuation share %v without a sequence start share", share)
 			}
+			// FIXME: it doesn't look like we check whether all the shares belong to the same namespace.
```
cristaloleg, line 72: This PR ?

cmwaters: Are you implying I should fix it in this PR?

cristaloleg: Nah, just asking. Thanks for filling the issues https://github.com/celestiaorg/go-square/issues/78

PR #80, opened 2024-06-19
feat: support encoding and decoding of v1 shares

cmwaters:
This adds the logic of encoding and decoding the signer bytes to the share (only for the first share in the sequence and only for v1 share versions).

Since we're encapsulating the types, many of the length checks can be removed and thus much data in the square does not need to return an error.


shares/parse.go
```diff
@@ -36,13 +36,7 @@ func ParseShares(shares []Share, ignorePadding bool) ([]ShareSequence, error) {
 	currentSequence := ShareSequence{}
 
 	for _, share := range shares {
-		if err := share.Validate(); err != nil {
```
cristaloleg, line 39: Why validation is dropped?

cmwaters: Shares are made immutable (we encapsulate the []byte in a private field) and validation is done during construction so validating shares every time we call a function on it is not necessary

vgonkivs (approved):
Looks good to me. One idea for further implementation: include signer in the commitment, so we can choose between 2 equal blobs posted by different signers

blob/blob.go
```diff
@@ -37,13 +40,31 @@ type Blob struct {
...
+		return nil, errors.New("share version 1 requires signer of size 20 bytes")
+	}
 	return &Blob{
 		namespace:    ns,
 		data:         data,
 		shareVersion: shareVersion,
 		signer:       signer,
+	}, nil
+}
+
```
rach-id, line 57: [not blocking]
is there a place with v0 and v1 blobs are documented? what do they contain and how they're constructed?

inclusion/commitment_test.go
```diff
@@ -81,6 +82,14 @@ func TestCreateCommitment(t *testing.T) {
 			expected:     []byte{0x31, 0xf5, 0x15, 0x6d, 0x5d, 0xb9, 0xa7, 0xf5, 0xb4, 0x3b, 0x29, 0x7a, 0x14, 0xc0, 0x70, 0xc2, 0xcc, 0x4e, 0xf3, 0xd6, 0x9d, 0x87, 0xed, 0x8, 0xad, 0xdd, 0x21, 0x6d, 0x9b, 0x9f, 0xa1, 0x18},
 			shareVersion: shares.ShareVersionZero,
 		},
+		{
+			name:         "blob of one share with signer succeeds",
+			namespace:    ns1,
+			blob:         bytes.Repeat([]byte{0xFF}, shares.AvailableBytesFromSparseShares(2)-blob.SignerSize),
+			expected:     []byte{0x88, 0x3c, 0x74, 0x6, 0x4e, 0x8e, 0x26, 0x27, 0xad, 0x58, 0x8, 0x38, 0x9f, 0x1f, 0x19, 0x24, 0x19, 0x4c, 0x1a, 0xe2, 0x3c, 0x7d, 0xf9, 0x62, 0xc8, 0xd5, 0x6d, 0xf0, 0x62, 0xa9, 0x2b, 0x2b},
```
rach-id, line 89: [not blocking]
how was this value generated? it would be great to have more information on how to generate the test values

internal/test/factory.go
```diff
@@ -38,13 +38,21 @@ func RandomBytes(size int) []byte {
 	return b
 }
 
-func GenerateBlobTxWithNamespace(namespaces []namespace.Namespace, blobSizes []int) []byte {
+func GenerateBlobTxWithNamespace(namespaces []namespace.Namespace, blobSizes []int, version uint8) []byte {
 	blobs := make([]*blob.Blob, len(blobSizes))
 	if len(namespaces) != len(blobSizes) {
 		panic("number of namespaces should match number of blob sizes")
 	}
+	var err error
```
rach-id, line 46: [nit]
why is this declared here? given that we panic if we find an error

shares/share_builder.go
```diff
@@ -168,6 +168,17 @@ func (b *Builder) WriteSequenceLen(sequenceLen uint32) error {
 	return nil
 }
 
+// WriteSigner writes the signer's information to the share.
+func (b *Builder) WriteSigner(signer []byte) {
+	// only write the signer if it is the first share and the share version is 1
+	if b == nil || !b.isFirstShare || b.shareVersion != ShareVersionOne {
+		return
+	}
+	// NOTE: we don't check whether previous data has already been expected
```
rach-id, line 177: [nit]
I guess this note should also be written in the function docs so that users can see it when they're using it, instead of having to read the code to know

shares/shares.go
```diff
@@ -14,70 +15,51 @@ type Share struct {
...
-	// the info byte is the first byte after the namespace
-	unparsed := s.data[namespace.NamespaceSize]
-	return ParseInfoByte(unparsed)
-}
-
 func NewShare(data []byte) (*Share, error) {
 	if err := validateSize(data); err != nil {
 		return nil, err
 	}
+	// TODO: we should also validate namespace
```
rach-id, line 22: should we create an issue for this?

shares/shares.go
```diff
@@ -14,70 +15,51 @@ type Share struct {
...
 	if len(data) != ShareSize {
 		return fmt.Errorf("share data must be %d bytes, got %d", ShareSize, len(data))
 	}
 	return nil
 }
 
-func (s *Share) Len() int {
-	return len(s.data)
+// Namespace returns the shares namespace
+// TODO: we could validate it the namespace in the constructor
```
rach-id, line 34:
```suggestion
// TODO: we could validate the namespace in the constructor
```

Also similar, do we need an issue for this?

shares/shares.go
```diff
@@ -90,33 +72,38 @@ func (s Share) IsCompactShare() (bool, error) {
 	return isCompact, nil
 }
 
+// GetSigner returns the signer of the share, if the
+// share is not of type v1 and is not the first share in a sequence
+// it returns nil
+func GetSigner(share Share) []byte {
+	infoByte := share.InfoByte()
+	if infoByte.Version() != ShareVersionOne {
+		return nil
```
rach-id, line 81: why not return an error?

shares/shares.go
```diff
@@ -90,33 +72,38 @@ func (s Share) IsCompactShare() (bool, error) {
...
+	}
+	if !infoByte.IsSequenceStart() {
+		return nil
```
rach-id, line 84: similar, why not error?

shares/shares.go
```diff
@@ -90,33 +72,38 @@ func (s Share) IsCompactShare() (bool, error) {
...
+	}
+	startIndex := namespace.NamespaceSize + ShareInfoBytes + SequenceLenBytes
+	endIndex := startIndex + blob.SignerSize
+	return share.data[startIndex:endIndex]
```
rach-id, line 88: doesn't it make sense to check whether the length of the `share.data` can contain the given range? to avoid any slice out of bounds issues

shares/shares.go
```diff
@@ -90,33 +72,38 @@ func (s Share) IsCompactShare() (bool, error) {
...
-	if !isSequenceStart {
-		return 0, nil
+func (s *Share) SequenceLen() uint32 {
+	if !s.IsSequenceStart() {
+		return 0
 	}
 
 	start := namespace.NamespaceSize + ShareInfoBytes
 	end := start + SequenceLenBytes
-	if len(s.data) < end {
```
rach-id, line 107: why was this check deleted? I think it's good to have to avoid out of bounds issues

blob/blob.go
cmwaters: Currently we just have this cip: https://github.com/celestiaorg/CIPs/blob/main/cips/cip-21.md

But it makes sense to definitely make sure the new blob type is clearly documented. I will need to think about where the best place for that is. We should probably have it mentioned in docs.celestia.org but perhaps we want to have a go.doc or readme.md in the code to explain it.

Will just jot this down as an issue so I don't forget

internal/test/factory.go
cmwaters: We need to declare it ahead of line 52 where it gets assigned

shares/shares.go
cmwaters: It's the same issue just mentioned in two different places. I will create the issue. Thanks for reminding

shares/share_builder.go
cmwaters: I think instead I want to make an assertion about the size of the `rawShareData` before making the append to know whether the other writes have happened. Personally, I'm not sure about the entire builder struct (I think it should at least be private)

blob/blob.go
```diff
@@ -25,9 +25,12 @@ const ProtoIndexWrapperTypeID = "INDX"
 // MaxShareVersion is the maximum value a share version can be. See: [shares.MaxShareVersion].
 const MaxShareVersion = 127
 
+// SignerSize is the size of the signer in bytes.
+const SignerSize = 20
+
 // Blob (stands for binary large object) is a core type that represents data
 // to be submitted to the Celestia network alongside an accompanying namespace
-// and optional signer (for proving the signer of the blob)
+// and optional signer (for proving the author of the blob)
```
rootulp, line 33: [nit] I think this was accidental b/c we renamed it to signer in a different PR

```suggestion
// and optional signer (for proving the signer of the blob)
```

blob/blob.go
```diff
@@ -37,13 +40,31 @@ type Blob struct {
 
 // New creates a new coretypes.Blob from the provided data after performing
 // basic stateless checks over it.
-func New(ns ns.Namespace, data []byte, shareVersion uint8, signer []byte) *Blob {
+func New(ns ns.Namespace, data []byte, shareVersion uint8, signer []byte) (*Blob, error) {
+	if shareVersion == 0 && signer != nil {
+		return nil, errors.New("share version 0 does not support signer")
+	}
+	if shareVersion == 1 && len(signer) != SignerSize {
+		return nil, errors.New("share version 1 requires signer of size 20 bytes")
```
rootulp, line 48: [optional] use the constant defined above

 ```suggestion
		return nil, fmt.Errorf("share version 1 requires signer of size %d bytes", SignerSize)
```

File: blob/blob.go (hunk 2 above)
Comment (rootulp) [reply blob/blob.go:57]: Note we have https://celestiaorg.github.io/celestia-app/specs/shares.html#share-version so it probably warrants a mention there (eventually)

File: blob/blob.go [hunk 14]
```diff
@@ -37,13 +40,31 @@ type Blob struct {
...
 		data:         data,
 		shareVersion: shareVersion,
 		signer:       signer,
+	}, nil
+}
+
+func NewV0(ns ns.Namespace, data []byte) *Blob {
+	blob, err := New(ns, data, 0, nil)
+	if err != nil {
+		panic(err)
```
rootulp, line 61: this panic kinda scares me. Can a malicious entity provide an invalid v0 namespace to an honest light node or consensus node and cause it to hit this panic? I think it is preferable to bubble up errors from low level code like this and let the application (celestia-app or celestia-node) decide what to do with the error if it encounters one.

shares/shares.go
```diff
@@ -90,33 +72,38 @@ func (s Share) IsCompactShare() (bool, error) {
...
+		return nil
+	}
+	startIndex := namespace.NamespaceSize + ShareInfoBytes + SequenceLenBytes
+	endIndex := startIndex + blob.SignerSize
+	return share.data[startIndex:endIndex]
+}
+
 // SequenceLen returns the sequence length of this *share and optionally an
 // error. It returns 0, nil if this is a continuation share (i.e. doesn't
 // contain a sequence length).
```
rootulp, line 93: godoc needs updating.

```suggestion
// SequenceLen returns the sequence length of this share. 
// It returns 0 if this is a continuation share because then it doesn't contain a sequence length.
```

rootulp (approved):

blob/blob.go
cmwaters: Good point, I'll add it to the issue

blob/blob.go
cmwaters: Currently, the panic is not possible, because the namespace is encapsulated so we know it's valid and there's not a signer that could be of an invalid length. However since there could be changes in the future that could make errors acceptable, I think you're right and I will switch this to being an error

cmwaters:
@rootulp I will address your nits in a follow up PR (just so as to not block this)

Issue #79, opened 2024-06-19
Merge namespace and blob packages into shares


cmwaters:
### Summary

After some discussion, it seems apparent that the types `Share`, `Blob` and `Namespace` are inextricably linked. The share package after all is an encoding protocol that converts `Blob`s into `Share`s and vice versa.

### Proposal

Merge the `blob` and `namespace` packages into the `share` package

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #82, opened 2024-06-19
chore!: merge blob pkg into share pkg

cmwaters:
Closes: https://github.com/celestiaorg/go-square/issues/79

The decision comes about because it's more intuitive that blobs reside in the share package which is responsible for converting blobs into shares and vice versa.


rach-id (approved):
Just one question, why is the package called `shares` and not `share`? all the other packages are in singular except for that one

cmwaters:
> Just one question, why is the package called `shares` and not `share`? all the other packages are in singular except for that one

I would agree. In fact I think it's common practice to use singular form (ref: https://go.dev/blog/package-names). I can change it to `share`. I was also thinking that something like `split` might be a better description and not lead to problems like `share := share.NewShare(); share.Version()`

rootulp:
> Just one question, why is the package called shares and not share? 

+1 I think it was accidentally called `shares` a long time ago (https://github.com/celestiaorg/celestia-app/pull/627) and we just haven't made the breaking change to rename it to singular `share`.

rootulp (approved):

Issue #81, opened 2024-06-19
Add benchmarks to share encoding / decoding


cmwaters:
### Summary

Since we are making modifications to the share package, we should also look to put some benchmarks so we can be aware of any performance regressions

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #83, opened 2024-06-19
test: add benchmarks to share encoding and decoding

cmwaters:
Closes: https://github.com/celestiaorg/go-square/issues/81

Note: need to also backport this to v1.x


cristaloleg (dismissed):
SGTM

internal/test/factory.go
```diff
@@ -77,6 +77,18 @@ func GenerateBlobTxs(numTxs, blobsPerPfb, blobSize int) [][]byte {
 	return txs
 }
 
+func GenerateBlobs(blobSizes ...int) []*blob.Blob {
+	blobs := make([]*blob.Blob, len(blobSizes))
+	var err error
+	for i, size := range blobSizes {
+		blobs[i], err = blob.New(namespace.RandomBlobNamespace(), RandomBytes(size), shares.ShareVersionZero, nil)
```
rootulp, line 84: [optional follow up]: it could be useful for tests to generate blobs of share version 0 and share version 1 so that we catch edge cases caused by the signer.

rootulp (approved):

internal/test/factory.go
cmwaters: I think for some tests I will want to go for table like testing where both v1 and v0 are convered

rootulp: SGTM can resolve.

rootulp (approved):

PR #84, opened 2024-06-19
chore: backport share benchmarks

cmwaters:
Ref: https://github.com/celestiaorg/go-square/issues/81

### DIFF
--- a/internal/test/factory.go
+++ b/internal/test/factory.go
@@ -71,12 +71,8 @@ func GenerateBlobTxs(numTxs, blobsPerPfb, blobSize int) [][]byte {
 
 func GenerateBlobs(blobSizes ...int) []*blob.Blob {
 	blobs := make([]*blob.Blob, len(blobSizes))
-	var err error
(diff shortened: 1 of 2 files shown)


internal/test/factory.go
```diff
@@ -69,6 +69,14 @@ func GenerateBlobTxs(numTxs, blobsPerPfb, blobSize int) [][]byte {
 	return txs
 }
 
+func GenerateBlobs(blobSizes ...int) []*blob.Blob {
```
ninabarbakadze, line 72: should we give this a godoc

ninabarbakadze (approved):
LGTM

internal/test/factory.go
cmwaters: It's in an `internal` director, so this wouldn't be public to users and I think it should be self explanatory enough to us core developers so forgive me for letting this one pass

rootulp (approved):

evan-forbes (approved):

Issue #86, opened 2024-06-20
Validate the namespace in the Share constructor


cmwaters:
_Originally posted by @cmwaters in https://github.com/celestiaorg/go-square/pull/80#discussion_r1647512377_

PR #87, opened 2024-06-21
feat: validate namespace in share constructor

cmwaters:
Closes: https://github.com/celestiaorg/go-square/issues/86

This PR also addresses the nits from @rootul's review [here](https://github.com/celestiaorg/go-square/pull/80#discussion_r1648021038)


rootulp (approved):

namespace/namespace.go
```diff
@@ -110,6 +94,14 @@ func (n Namespace) ID() []byte {
 	return n.data[NamespaceVersionSize:]
 }
 
+func Validate(version uint8, id []byte) error {
```
Wondertan, line 97: Needs a little godoc.

namespace/namespace.go
```diff
@@ -46,17 +40,7 @@ func MustNew(version uint8, id []byte) Namespace {
...
-	}
-
-	err := validateVersionSupported(bytes[VersionIndex])
-	if err != nil {
-		return Namespace{}, err
-	}
-
-	err = validateID(bytes[VersionIndex], bytes[NamespaceVersionSize:])
-	if err != nil {
+	if err := Validate(bytes[VersionIndex], bytes[NamespaceVersionSize:]); err != nil {
```
Wondertan, line 43: nit: to avoid confusing byte slicing you could do:
```go
ns := Namespace{data: bytes}
err := Validate(ns.Version(), ns.ID())
if err != nil {
   return Namespace{}, err
}

return ns
```

Wondertan: Still, I'm not sure why we wouldn't add Validate as a method. (I know it's guaranteed to be valid after the construction but still)

shares/blob.go
```diff
@@ -39,12 +42,8 @@ func NewBlob(ns ns.Namespace, data []byte, shareVersion uint8, signer []byte) (*
 	}, nil
 }
 
-func NewV0Blob(ns ns.Namespace, data []byte) *Blob {
-	blob, err := NewBlob(ns, data, 0, nil)
-	if err != nil {
-		panic(err)
-	}
-	return blob
+func NewV0Blob(ns ns.Namespace, data []byte) (*Blob, error) {
```
Wondertan, line 45: godoc

namespace/namespace.go
cmwaters: It gets used by the `share` package

Wondertan: :merge-it:

cmwaters:
I plan to merge the namespace package into the shares package. This will mean that `Validate` no longer needs to be a public function

Issue #88, opened 2024-06-24
Change shares package name to share


cmwaters:
> Just one question, why is the package called shares and not share? 

+1 I think it was accidentally called `shares` a long time ago (https://github.com/celestiaorg/celestia-app/pull/627) and we just haven't made the breaking change to rename it to singular `share`.

_Originally posted by @rootulp in https://github.com/celestiaorg/go-square/issues/82#issuecomment-2183190907_

PR #90, opened 2024-06-27
chore!: rename shares to share

cmwaters:
Closes: https://github.com/celestiaorg/go-square/issues/88


README.md
```diff
@@ -10,7 +10,7 @@ blob      | Package blob provides types and functions for working with blobs, bl
 inclusion | Package inclusion contains functions to generate the blob share commitment from a given blob.
 merkle    | Package merkle computes a deterministic minimal height Merkle tree hash.
 namespace | Package namespace contains the Namespace data structure.
-shares    | Package shares contains the Share data structure.
+shares    | package share contains the Share data structure.
```
rootulp, line 13:
```suggestion
share     | Package share contains the Share data structure.
```

share/blob.pb.go
```diff
@@ -4,7 +4,7 @@
 // 	protoc        (unknown)
 // source: shares/blob.proto
```
rootulp, line 5: I think this prob needs a `make proto-gen` so that this comment gets updated

share/shares.go
```diff
@@ -1,5 +1,5 @@
-// Package shares contains the Share data structure.
-package shares
+// package share contains the Share data structure.
```
rootulp, line 1: [optional][follow-up] we could expand this to:

```suggestion
// Package share contains the data structures: share, namespace, blob.
```

rootulp (approved):
> exported: type name will be used as share.ShareSequence by other packages, and that stutters; consider calling this Sequence (revive)

I agree with this suggestion but we def don't have to do it in this PR. IMO optional if we want to include in the next breaking release

README.md
cmwaters: I should also remove the other packages that no longer exist. Thanks for pointing it out

cmwaters:
> I agree with this suggestion but we def don't have to do it in this PR. IMO optional if we want to include in the next breaking release

I think golangci-lint will complain until this is fixed

rootulp:
We could add a golangci-lint ignore directive if you want to merge this first and do a separate PR for the rename from `ShareSequence` -> `Sequence`

evan-forbes (approved):

README.md
```diff
@@ -6,11 +6,9 @@
 
 Package   | Description
 ----------|---------------------------------------------------------------------------------------------------------------------
-blob      | Package blob provides types and functions for working with blobs, blob transactions, and index wrapper transactions.
 inclusion | Package inclusion contains functions to generate the blob share commitment from a given blob.
 merkle    | Package merkle computes a deterministic minimal height Merkle tree hash.
-namespace | Package namespace contains the Namespace data structure.
-shares    | package share contains the Share data structure.
+share    | package share contains the Share data structure.
```
rootulp, line 11:
```suggestion
share    | Package share contains the Share data structure.
```

rootulp (approved):

rach-id (approved):

cristaloleg:
Share is a very good name for a variable. Just saying. 

Reminds me net/url pain.

evan-forbes (approved):

cmwaters:
> Share is a very good name for a variable. Just saying.
> 
> Reminds me net/url pain.

You're totally right. I had brought it up before when I wanted to call it something like `enc` for encoding or `split`

Issue #39, opened 2024-02-28
This repo shouldn't maintain Merkle tree implemenation


Wondertan:
`merkle` pkg contains Merkle tree extracted from Comet/Tendermint. This is a pretty standart RFC compatible Merkle tree implementation that should not be part of this repo. We should consider alternative and likely more efficient implementations out there, as depending on CometBFT is not an option for this repository.

rootulp:
One idea that came up in a synchronous call is we could potentially make the Merkle tree in celestia-core a unique Go package and export it for consumption here so that we can delete the duplicate copy in this repo.

cristaloleg:
This! For a new Celestia member for me it also was not that obvious why we've https://github.com/celestiaorg/smt (which is archived), why we've MT in `go-square` (and partially in `celestia-core`).

PR #91, opened 2024-06-27
chore!: remove merkle library

cmwaters:
This reverts commit fdf1e920e58a838598c959315a5c50b7f0ee54ce.

Closes: https://github.com/celestiaorg/go-square/issues/39


Wondertan:
Shoot, we haven't decided what we do instead.

cmwaters:
> Shoot, we haven't decided what we do instead.

We're actually not using this library atm (we're still depending on the `merkle` library in `celestia-core`). I don't mind if we decide to spin out it's own repo, to make `merkle` it's own go.mod, or just leave it as is

rootulp:
Note: there are usages of go-square/merkle in celestia-app (e.g. [here](https://github.com/celestiaorg/celestia-app/blob/941538c20bf30a62e3753a152bf87302f62ce529/pkg/da/data_availability_header.go#L9)) so I'm guessing those usages will need to be updated to celestia-core `merkle` instead when celestia-app bumps to go-square v2.

rootulp (approved):
Seems fine to me. In go-square v2 release notes we should provide guidance to any consumers who use go-square/merkle to migrate to something else (potentially celestia-core merkle)

cristaloleg (approved):
yay

Wondertan (changes requested):
> Actually I'm wrong. We do use it so would have to revert back to celestia-core

Just saw the edit. I am strictly against go-square depending on celestia-core, even if the merkle pkg is a separate go.mod inside

rootulp:
go-square doesn't need to depend on celestia-core b/c nothing breaks if we remove the `merkle` package in this repo.

Wondertan:
> We do use it so would have to revert back to celestia-core

What does revert back mean?

rootulp:
I think consumers of go-square/merkle would have to revert back to consuming it from celestia-core.

Wondertan (approved):
I see, I misunderstood the message then

PR #92, opened 2024-06-27
chore: internalize unneeded go apis

cmwaters:
This PR does an initial pass of the `share` repo and internalizes things like test methods or the `Builder` struct that should not be public.


share/share_builder.go
```diff
@@ -5,23 +5,23 @@ import (
...
 
-func NewEmptyBuilder() *Builder {
-	return &Builder{
+func newEmptyBuilder() *builder {
+	return &builder{
 		rawShareData: make([]byte, 0, ShareSize),
 	}
 }
 
 // NewBuilder returns a new share builder.
```
rootulp, line 22:
```suggestion
// newBuilder returns a new share builder.
```

rootulp (dismissed):

rootulp (approved):

Issue #77, opened 2024-06-18
Pre-verify blobs once instead of twice


evan-forbes:
converting this todo to an issue

```
// TODO: we should look at having a go native BlobTx type
			// that we have already verified instead of doing it twice here
			```

_Originally posted by @evan-forbes in https://github.com/celestiaorg/go-square/pull/74#discussion_r1645153644_

Issue #33, opened 2024-02-19
Consider renaming blob proto package


rootulp:
## Context

https://github.com/celestiaorg/go-square/blob/4135f359b20ddd97b6cb74ecc831f70ffb73f8a5/blob/blob.proto#L2

## Problem

```
Files with package "pkg.blob" must be within a directory "pkg/blob" relative to root but were in directory "blob". (PACKAGE_DIRECTORY_MATCH)
```

## Proposal

Resolve the warning

PR #94, opened 2024-07-18
chore!: add native blob tx

cmwaters:
Closes: https://github.com/celestiaorg/go-square/issues/77, https://github.com/celestiaorg/go-square/issues/33


share/blob_tx.go
```diff
@@ -0,0 +1,78 @@
...
+	ProtoBlobTxTypeID = "BLOB"
+)
+
+type BlobTx struct {
+	Tx    []byte
+	Blobs []*Blob
+}
+
+// UnmarshalBlobTx attempts to unmarshal a transaction into blob transaction. If an
+// error is thrown, false is returned.
```
rootulp, line 22: IMO the second sentence is a bit misleading b/c this function can return an error and isBlob=false

```suggestion
// UnmarshalBlobTx attempts to unmarshal a transaction into blob transaction.
```

share/blob_tx.go
```diff
@@ -0,0 +1,78 @@
...
+func UnmarshalBlobTx(tx []byte) (*BlobTx, bool, error) {
```
rootulp, line 23: [nit] I think providing a name for the second return parameter would help communicate it's meaning

```suggestion
func UnmarshalBlobTx(tx []byte) (blobTx *BlobTx, isBlobTx bool, err error) {
```

share/utils.go
```diff
@@ -40,7 +40,7 @@ func zeroPadIfNecessary(share []byte, width int) (padded []byte, bytesOfPadding
 // in compact shares where units (i.e. a transaction) are prefixed with a length
 // delimiter that is encoded as a varint. Input should not contain the namespace
 // ID or info byte of a share.
-func ParseDelimiter(input []byte) (inputWithoutLenDelimiter []byte, unitLen uint64, err error) {
+func parseDelimiter(input []byte) (inputWithoutLenDelimiter []byte, unitLen uint64, err error) {
```
rootulp, line 43: [nit]

go doc `ParseDelimiter` => `parseDelimiter`

rootulp (approved):
I'm a fan of distinct files for blob_tx.go and index_wrapper.go. Overall I'm a fan of one file per significant struct (i.e. `type BlobTx struct`)

share/blob_tx.go
cmwaters, line 22:
```suggestion
// UnmarshalBlobTx attempts to unmarshal a transaction into blob transaction. It returns a boolean 
// If the bytes are of type BlobTx and an error if there is a problem with decoding
```

share/blob_tx.go
cmwaters: I'm just going to try make it clear in the comments

rootulp (approved):

builder.go
```diff
@@ -42,14 +43,17 @@ func NewBuilder(maxSquareSize int, subtreeRootThreshold int, txs ...[]byte) (*Bu
...
+		Pfbs:                 make([]*v1.IndexWrapper, 0),
 		Txs:                  make([][]byte, 0),
 		TxCounter:            share.NewCompactShareCounter(),
 		PfbCounter:           share.NewCompactShareCounter(),
 	}
 	seenFirstBlobTx := false
 	for idx, tx := range txs {
-		blobTx, isBlobTx := share.UnmarshalBlobTx(tx)
+		blobTx, isBlobTx, err := share.UnmarshalBlobTx(tx)
+		if err != nil && isBlobTx {
```
rach-id, line 54: what if `err !=nil` but `isBlobTx == false`, where does the unmarshalling error get returned?

share/blob_tx.go
```diff
@@ -0,0 +1,78 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+const (
+	// ProtoBlobTxTypeID is included in each encoded BlobTx to help prevent
+	// decoding binaries that are not actually BlobTxs.
+	ProtoBlobTxTypeID = "BLOB"
+)
+
+type BlobTx struct {
```
Wondertan, line 16: Is the idea of making go-square hold actual transactions? Isn't that a concern of the app?

evan-forbes (approved):

PR #97, opened 2024-07-24
chore!: break out BlobTx and IndexWrapper types into separate tx package

cmwaters:
While updating `celestia-app`, it felt out of place that BlobTx and IndexWrapper should be part of the `shares` package. This moves them out


README.md
```diff
@@ -7,8 +7,10 @@
 Package   | Description
 ----------|---------------------------------------------------------------------------------------------------------------------
 inclusion | Package inclusion contains functions to generate the blob share commitment from a given blob.
+proto     | Package contains proto definitions and go generated code
```
rootulp, line 10: [question] does this proto package actually exist? It looks like the proto definition for BlobTx and IndexWrapper are still inside proto/blob/v1

rootulp (approved):
Do we also want to move the proto definitions for BlobTx and IndexWrapper from `proto/blob/v1/blob.proto` to something like `proto/tx/blobtx.proto` and `proto/tx/indexwrapper.proto`

README.md
rootulp: put another way: there is no Go package named `proto`

rach-id (approved):

README.md
cmwaters: No but here is generated code in the proto directory and that's what I was referring to. I'm happy to remove it

cmwaters:
> Do we also want to move the proto definitions for BlobTx and IndexWrapper from `proto/blob/v1/blob.proto` to something like `proto/tx/blobtx.proto` and `proto/tx/indexwrapper.proto`

Yeah it seems more logical to separate them but we don't need to

README.md
rootulp: I thought this list was just Go packages but I can see an argument for expanding this list to all directories and not just Go packages

PR #98, opened 2024-07-26
fix: share shouldn't validate namespace versions

cmwaters:
Whilst updating `celestia-app`, I realised that we falsely verify the namespace of the shares using the criteria that it must be a "usable namespace". This validation shouldn't be here because the protocol uses arbitrary namespaces when erasure coding. This resolves the problem

### DIFF
--- a/share/namespace.go
+++ b/share/namespace.go
@@ -94,7 +94,7 @@ func (n Namespace) ID() []byte {
 
 // ValidateUserNamespace returns an error if the provided version is not
 // supported or the provided id does not meet the requirements
-// for the provided version. This should be user for validating
+// for the provided version. This should be used for validating
 // user specified namespaces
 func ValidateUserNamespace(version uint8, id []byte) error {
 	err := validateVersionSupported(version)
--- a/share/namespace.go
+++ b/share/namespace.go
@@ -9,9 +9,10 @@ type Namespace struct {
 	data []byte
 }
 
-// NewNamespace returns a new namespace with the provided version and id.
+// NewNamespace validates the provided version and id and returns a new namespace.
+// This should be used for user specified namespaces.
 func NewNamespace(version uint8, id []byte) (Namespace, error) {
-	if err := validate(version, id); err != nil {
+	if err := IsValidNamespace(version, id); err != nil {
 		return Namespace{}, err
 	}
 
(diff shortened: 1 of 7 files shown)


ninabarbakadze (dismissed):

share/namespace.go
```diff
@@ -90,7 +92,11 @@ func (n Namespace) ID() []byte {
 	return n.data[NamespaceVersionSize:]
 }
 
-func validate(version uint8, id []byte) error {
+// IsValidNamespace returns an error if the provided version is not
+// supported or the provided id does not meet the requirements
+// for the provided version. This should be user for validating
+// user specified namespaces
+func IsValidNamespace(version uint8, id []byte) error {
```
cristaloleg, line 99: My 2c: `IsXXX` or `IsValidXXX` functions mostly returns `bool`, but when the returned param is `error` `ValidateXXX` is more common.

cmwaters: Fair point

evan-forbes:
> This validation shouldn't be here because the protocol uses arbitrary namespaces when erasure coding.

how was this avoided before? is getting rid of this check consensus breaking?

share/namespace.go
cristaloleg: Not sure do we need `User` in the func name. I might missing some context, so feel free to resolve the comment.

cmwaters:
> how was this avoided before? is getting rid of this check consensus breaking?

We didn't do it before. It's in the constructor. I added it in because I thought it should be there but it shouldn't

rach-id (commented):
> This validation shouldn't be here because the protocol uses arbitrary namespaces when erasure coding. This resolves the problem

what do you mean by arbitrary namespaces?

cmwaters:
> what do you mean by arbitrary namespaces?

When it erasure codes the share, we get erasure coded namespaces which make use any version and has no restrictions

share/namespace.go
```diff
@@ -90,7 +92,11 @@ func (n Namespace) ID() []byte {
 	return n.data[NamespaceVersionSize:]
 }
 
-func validate(version uint8, id []byte) error {
+// ValidateUserNamespace returns an error if the provided version is not
+// supported or the provided id does not meet the requirements
+// for the provided version. This should be user for validating
```
rootulp, line 97:
```suggestion
// for the provided version. This should be used for validating
```

share/share.go
```diff
@@ -54,7 +55,7 @@ func (s *Share) Version() uint8 {
 }
 
 // DoesSupportVersions checks if the share version is supported
```
rootulp, line 57:
```suggestion
// doesSupportVersions checks if the share version is supported
```

rootulp (dismissed):

rootulp (approved):

ninabarbakadze (approved):

PR #101, opened 2024-08-01
chore: fix share and blob validation

cmwaters:
I've now integrated celestia-app locally. These are the last set of changes in order to make everything work. These are:
- Remove validation of share verision. Erasure coded data doesn't comply with this rule. The validation has been added to the parsing function
- Check for empty namespaces in the blob constructor. There's a bit of a footgun here where users could set an empty namespace.

### DIFF
--- a/share/blob.go
+++ b/share/blob.go
@@ -25,6 +25,9 @@ func NewBlob(ns Namespace, data []byte, shareVersion uint8, signer []byte) (*Blo
 	if len(data) == 0 {
 		return nil, errors.New("data can not be empty")
 	}
+	if ns.IsEmpty() {
+		return nil, errors.New("namespace can not be empty")
+	}
 	if shareVersion == 0 && signer != nil {
 		return nil, errors.New("share version 0 does not support signer")
 	}
@@ -79,8 +82,8 @@ func NewBlobFromProto(pb *v1.BlobProto) (*Blob, error) {
 	if pb.NamespaceVersion > NamespaceVersionMax {
 		return nil, errors.New("namespace version can not be greater than MaxNamespaceVersion")
 	}
-	if len(pb.Data) == 0 {
-		return nil, errors.New("blob data can not be empty")
+	if pb.ShareVersion > MaxShareVersion {
+		return nil, fmt.Errorf("share version can not be greater than MaxShareVersion %d", MaxShareVersion)
 	}
 	ns, err := NewNamespace(uint8(pb.NamespaceVersion), pb.NamespaceId)
 	if err != nil {
@@ -122,6 +125,12 @@ func (b *Blob) DataLen() int {
 // Compare is used to order two blobs based on their namespace
 func (b *Blob) Compare(other *Blob) int {
 	return b.namespace.Compare(other.namespace)
+}
+
+// IsEmpty returns true if the blob is empty. This is an invalid
+// construction that can only occur if using the nil value
+func (b *Blob) IsEmpty() bool {
+	return b.namespace.IsEmpty()
 }
 
 // Sort sorts the blobs by their namespace.
--- a/share/blob.go
+++ b/share/blob.go
@@ -136,9 +136,11 @@ func (b *Blob) Compare(other *Blob) int {
 }
 
 // IsEmpty returns true if the blob is empty. This is an invalid
-// construction that can only occur if using the nil value
+// construction that can only occur if using the nil value. We
+// only check that the data is empty but this also implies that
+// all other fields would have their zero value
 func (b *Blob) IsEmpty() bool {
-	return b.namespace.IsEmpty()
+	return len(b.data) == 0
 }
 
 // Sort sorts the blobs by their namespace.
--- a/share/blob.go
+++ b/share/blob.go
@@ -28,14 +28,22 @@ func NewBlob(ns Namespace, data []byte, shareVersion uint8, signer []byte) (*Blo
 	if ns.IsEmpty() {
 		return nil, errors.New("namespace can not be empty")
 	}
-	if shareVersion == 0 && signer != nil {
-		return nil, errors.New("share version 0 does not support signer")
+	if ns.Version() != NamespaceVersionZero {
+		return nil, fmt.Errorf("namespace version must be %d got %d", NamespaceVersionZero, ns.Version())
 	}
-	if shareVersion == 1 && len(signer) != SignerSize {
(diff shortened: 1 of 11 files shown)


share/blob.go
```diff
@@ -124,6 +127,12 @@ func (b *Blob) Compare(other *Blob) int {
 	return b.namespace.Compare(other.namespace)
 }
 
+// IsEmpty returns true if the blob is empty. This is an invalid
```
rootulp, line 130: I would expect `blob.IsEmpty` to check if the data inside Blob was empty. Since this checks if the namespace is Empty, perhaps:

```suggestion
// IsEmpty returns true if the blob's namespace is empty. This is an invalid
```

rootulp (approved):

tx/blob_tx.go
```diff
@@ -59,7 +59,7 @@ func MarshalBlobTx(tx []byte, blobs ...*share.Blob) ([]byte, error) {
 	}
 	// nil check
 	for i, b := range blobs {
-		if b == nil {
+		if b == nil || b.IsEmpty() {
```
staheri14, line 62: [nit] The name `IsEmpty` does not immediately suggest that it checks the emptiness of the namespace. I suggest either renaming the method or adding a clarifying comment here to avoid confusion. Otherwise, one might intuitively think that it checks the emptiness of blobs rather than the namespace.

share/blob.go
```diff
@@ -25,6 +25,9 @@ func NewBlob(ns Namespace, data []byte, shareVersion uint8, signer []byte) (*Blo
 	if len(data) == 0 {
 		return nil, errors.New("data can not be empty")
 	}
+	if ns.IsEmpty() {
```
staheri14, line 28: Now that we are enhancing the checks for namespace, wondering shouldn't we consider checking that namespace version is `0` as per [specs](https://celestiaorg.github.io/celestia-app/specs/namespace.html#version-0) (it seems there is no other supported versions) or otherwise the namespace is one of the [reserved namespaces](https://celestiaorg.github.io/celestia-app/specs/namespace.html#reserved-namespaces)?

share/blob.go
```diff
@@ -79,8 +82,8 @@ func NewBlobFromProto(pb *v1.BlobProto) (*Blob, error) {
 	if pb.NamespaceVersion > NamespaceVersionMax {
 		return nil, errors.New("namespace version can not be greater than MaxNamespaceVersion")
 	}
-	if len(pb.Data) == 0 {
-		return nil, errors.New("blob data can not be empty")
+	if pb.ShareVersion > MaxShareVersion {
```
staheri14, line 85: I think we could also check against `SupportedShareVersions`?

staheri14 (commented):
LGTM, left some comments.

share/blob.go
cmwaters: It checks the namespace as proxy for the entire blob. By construction you can't have a blob with empty data and a non-empty namespace. I could have checked that all the contents of the blob had their "null" value but only one should suffice

tx/blob_tx.go
cmwaters: It's not about checking the emptiness of the namespace but of the entire blob. It checks the namespace as proxy for the entire blob. By construction you can't have a blob with empty data and a non-empty namespace. I could have checked that all the contents of the blob had their "null" value but only one should suffice

share/blob.go
cmwaters: If we wanted to do that we should add it in `NewBlob` rather than here

share/blob.go
cmwaters: Makes sense. Then we don't have to remember to do it outside this function

share/blob.go
Wondertan: I propose then doing `blob.Namespace().IsEmpty()` instead of proxying to avoid confusion.

share/blob.go
```diff
@@ -25,14 +25,25 @@ func NewBlob(ns Namespace, data []byte, shareVersion uint8, signer []byte) (*Blo
...
+			return nil, errors.New("share version 0 does not support signer")
+		}
+	case ShareVersionOne:
+		if len(signer) != SignerSize {
+			return nil, fmt.Errorf("share version 1 requires signer of size %d bytes", SignerSize)
+		}
+	// Note that we don't specifically check that shareVersion is less than 128 as this is caught
+	// by the default case
+	default:
+		return nil, fmt.Errorf("share version %d not supported. Please use 0 or 1", shareVersion)
```
Wondertan, line 46: If the goal is to make the general constructor more restrictive by adding share version check, I propose to make it private in favour of keeping  `NewBlobV0/1` public only. It a nit but:
* It's less code and less test cases to cover.
* Its scales better as adding new versions requires less cognitive load to think where to add new cases for new versions. You simply add a new versioned constructor and that's it.

Wondertan: You could also move namespace version checks to versioned constructors as that would simplify expressing relationships between namespace and share versions. E.g. BlobV3 does not work with NamespaceV0, but BlobV2 does.

tx/blob_tx.go
Wondertan: > By construction you can't have a blob with empty data and a non-empty namespace.

That should be documented on the Blob method. Or you could still do the nil check for readability, even though its not possible.

staheri14 (approved):
Thanks for addressing the comments, LGTM!

rootulp (approved):

share/blob.go
cmwaters: hmm I want to avoid this matrix of share and namespace versions

Wondertan: The version matrix is irrelevant mostly. The point is about removing unnecessary checks for something that we already enforce with versioned constructors. I dont see much value in doing the same thing twice.

The matrix point is an additional thought, that in case it happens(i presumed so) its easier to handle in versioned constructor rather then general.

tx/blob_tx.go
cmwaters: Do you want me to also add a check that blob.data is not length 0?

Wondertan: Zero data check would be clearer for the code readability than cognitive overhead of understanding why its only a namespace check

share/blob.go
cmwaters: We use `NewBlob` often when parsing from shares or from proto. If I were to make this private then I would then need to move the switch statements to these locations. I'm going to keep the `NewBlob` constructor as is

Wondertan (approved):

PR #104, opened 2024-08-10
don't merge this

S1nus:
## Overview
Solidity doesn't have dynamic memory arrays, so we tried to figure out how to compute the array size. Here it is.


rootulp:
Seems stale. Feel free to re-open if this is still needed.

PR #108, opened 2024-10-07
chore: expose generate subtree roots method (#106)

rach-id:
This PR exposes a method to generate the subtree roots. It will be handy in downstream repos (will be using it in Celestia-node) instead of re-implementing it.

cherry picked from commit 662d17db20089f0fcba6ef4369a49fb87e3b4127


cristaloleg (approved):
SGTM

PR #109, opened 2024-10-07
feat: add missing utilities used by celestia-node

vgonkivs:
## Overview
Added missing functionality that is commonly used by the node team.

### DIFF
--- a/share/blob.go
+++ b/share/blob.go
@@ -150,7 +150,7 @@ func SortBlobs(blobs []*Blob) {
 	})
 }
 
-// ToShares converts blob's data back to share.
+// ToShares converts blob's data back to shares.
 func (b *Blob) ToShares() ([]Share, error) {
 	splitter := NewSparseShareSplitter()
 	err := splitter.Write(b)
--- a/share/blob.go
+++ b/share/blob.go
@@ -151,6 +151,16 @@ func SortBlobs(blobs []*Blob) {
 	})
 }
 
+// ToShares converts blob's data back to share.
+func (b *Blob) ToShares() ([]Share, error) {
+	splitter := NewSparseShareSplitter()
+	err := splitter.Write(b)
+	if err != nil {
+		return nil, err
+	}
+	return splitter.Export(), nil
+}
+
 // IsValidBlobNamespace returns a true if this namespace is a valid user-specifiable
 // blob namespace.
 func IsValidBlobNamespace(ns Namespace) bool {
--- a/share/blob.go
+++ b/share/blob.go
@@ -151,9 +151,9 @@ func SortBlobs(blobs []*Blob) {
 	})
 }
 
-// IsBlobNamespace returns a true if this namespace is a valid user-specifiable
+// IsValidBlobNamespace returns a true if this namespace is a valid user-specifiable
 // blob namespace.
-func IsBlobNamespace(ns Namespace) bool {
+func IsValidBlobNamespace(ns Namespace) bool {
 	if ns.IsReserved() {
 		return false
 	}
--- a/share/blob.go
+++ b/share/blob.go
@@ -3,7 +3,6 @@ package share
 import (
 	"errors"
 	"fmt"
-	"slices"
 	"sort"
 
 	v1 "github.com/celestiaorg/go-square/v2/proto/blob/v1"
@@ -159,21 +158,4 @@ func (b *Blob) ToShares() ([]Share, error) {
 		return nil, err
 	}
 	return splitter.Export(), nil
-}
-
-// IsValidBlobNamespace returns a true if this namespace is a valid user-specifiable
-// blob namespace.
-func IsValidBlobNamespace(ns Namespace) bool {
-	if ns.IsReserved() {
-		return false
-	}
-
-	if !ns.IsUsableNamespace() {
-		return false
-	}
-
-	if !slices.Contains(SupportedBlobNamespaceVersions, ns.Version()) {
-		return false
-	}
-	return true
 }
--- a/share/blob.go
+++ b/share/blob.go
@@ -3,6 +3,7 @@ package share
 import (
 	"errors"
 	"fmt"
+	"slices"
 	"sort"
 
 	v1 "github.com/celestiaorg/go-square/v2/proto/blob/v1"
@@ -148,4 +149,21 @@ func SortBlobs(blobs []*Blob) {
 	sort.SliceStable(blobs, func(i, j int) bool {
 		return blobs[i].Compare(blobs[j]) < 0
 	})
+}
+
+// IsBlobNamespace returns a true if this namespace is a valid user-specifiable
+// blob namespace.
+func IsBlobNamespace(ns Namespace) bool {
+	if ns.IsReserved() {
+		return false
+	}
+
+	if !ns.IsUsableNamespace() {
+		return false
+	}
+
+	if !slices.Contains(SupportedBlobNamespaceVersions, ns.Version()) {
+		return false
+	}
+	return true
 }
--- a/share/blob_test.go
+++ b/share/blob_test.go
@@ -63,6 +63,16 @@ func TestBlobConstructor(t *testing.T) {
 	_, err = NewBlob(ns2, data, 0, nil)
 	require.Error(t, err)
 	require.Contains(t, err.Error(), "namespace version must be 0")
+
+	blob, err := NewBlob(ns, data, 0, nil)
+	require.NoError(t, err)
+	shares, err := blob.ToShares()
+	require.NoError(t, err)
+	blobList, err := parseSparseShares(shares)
+	require.NoError(t, err)
+	require.Len(t, blobList, 1)
+	require.Equal(t, blob, blobList[0])
+
 }
 
 func TestNewBlobFromProto(t *testing.T) {
--- a/share/blob_test.go
+++ b/share/blob_test.go
@@ -72,7 +72,6 @@ func TestBlobConstructor(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, blobList, 1)
 	require.Equal(t, blob, blobList[0])
-
 }
 
 func TestNewBlobFromProto(t *testing.T) {
--- a/share/namespace.go
+++ b/share/namespace.go
@@ -218,10 +218,6 @@ func (n Namespace) IsBlobNamespace() bool {
 		return false
 	}
 
-	if n.IsReserved() {
-		return false
-	}
-
 	if !slices.Contains(SupportedBlobNamespaceVersions, n.Version()) {
 		return false
 	}
--- a/share/namespace.go
+++ b/share/namespace.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"slices"
 )
 
 type Namespace struct {
@@ -34,11 +35,8 @@ func (n *Namespace) UnmarshalJSON(data []byte) error {
 // NewNamespace validates the provided version and id and returns a new namespace.
 // This should be used for user specified namespaces.
 func NewNamespace(version uint8, id []byte) (Namespace, error) {
-	if err := ValidateUserNamespace(version, id); err != nil {
-		return Namespace{}, err
-	}
-
-	return newNamespace(version, id), nil
+	ns := newNamespace(version, id)
+	return ns, ns.ValidateUserNamespace()
 }
 
 func newNamespace(version uint8, id []byte) Namespace {
@@ -66,13 +64,9 @@ func NewNamespaceFromBytes(bytes []byte) (Namespace, error) {
 	if len(bytes) != NamespaceSize {
 		return Namespace{}, fmt.Errorf("invalid namespace length: %d. Must be %d bytes", len(bytes), NamespaceSize)
 	}
-	if err := ValidateUserNamespace(bytes[VersionIndex], bytes[NamespaceVersionSize:]); err != nil {
-		return Namespace{}, err
-	}
 
-	return Namespace{
-		data: bytes,
-	}, nil
+	ns := Namespace{data: bytes}
+	return ns, ns.ValidateUserNamespace()
 }
 
 // NewV0Namespace returns a new namespace with version 0 and the provided subID. subID
@@ -123,17 +117,17 @@ func (n Namespace) String() string {
 // supported or the provided id does not meet the requirements
 // for the provided version. This should be used for validating
 // user specified namespaces
-func ValidateUserNamespace(version uint8, id []byte) error {
-	err := validateVersionSupported(version)
+func (n Namespace) ValidateUserNamespace() error {
+	err := n.validateVersionSupported()
 	if err != nil {
 		return err
 	}
-	return validateID(version, id)
+	return n.validateID()
 }
 
 // ValidateForData checks if the Namespace is of real/useful data.
-func ValidateForData(n Namespace) error {
-	if err := ValidateUserNamespace(n.Version(), n.ID()); err != nil {
+func (n Namespace) ValidateForData() error {
+	if err := n.ValidateUserNamespace(); err != nil {
 		return err
 	}
 	if !n.IsUsableNamespace() {
@@ -143,22 +137,22 @@ func ValidateForData(n Namespace) error {
 }
 
 // validateVersionSupported returns an error if the version is not supported.
-func validateVersionSupported(version uint8) error {
-	if version != NamespaceVersionZero && version != NamespaceVersionMax {
-		return fmt.Errorf("unsupported namespace version %v", version)
+func (n Namespace) validateVersionSupported() error {
+	if n.Version() != NamespaceVersionZero && n.Version() != NamespaceVersionMax {
+		return fmt.Errorf("unsupported namespace version %v", n.Version())
 	}
 	return nil
 }
 
 // validateID returns an error if the provided id does not meet the requirements
 // for the provided version.
-func validateID(version uint8, id []byte) error {
-	if len(id) != NamespaceIDSize {
-		return fmt.Errorf("unsupported namespace id length: id %v must be %v bytes but it was %v bytes", id, NamespaceIDSize, len(id))
+func (n Namespace) validateID() error {
+	if len(n.ID()) != NamespaceIDSize {
+		return fmt.Errorf("unsupported namespace id length: id %v must be %v bytes but it was %v bytes", n.ID(), NamespaceIDSize, len(n.ID()))
 	}
 
-	if version == NamespaceVersionZero && !bytes.HasPrefix(id, NamespaceVersionZeroPrefix) {
-		return fmt.Errorf("unsupported namespace id with version %v. ID %v must start with %v leading zeros", version, id, len(NamespaceVersionZeroPrefix))
+	if n.Version() == NamespaceVersionZero && !bytes.HasPrefix(n.ID(), NamespaceVersionZeroPrefix) {
+		return fmt.Errorf("unsupported namespace id with version %v. ID %v must start with %v leading zeros", n.Version(), n.ID(), len(NamespaceVersionZeroPrefix))
 	}
 	return nil
 }
@@ -215,6 +209,23 @@ func (n Namespace) Repeat(times int) []Namespace {
 		ns[i] = n.deepCopy()
 	}
 	return ns
+}
+
+// IsBlobNamespace returns true if this namespace is a valid user-specifiable
+// blob namespace.
+func (n Namespace) IsBlobNamespace() bool {
+	if err := n.ValidateForData(); err != nil {
+		return false
+	}
+
+	if n.IsReserved() {
+		return false
+	}
+
+	if !slices.Contains(SupportedBlobNamespaceVersions, n.Version()) {
+		return false
+	}
+	return true
 }
 
 func (n Namespace) Equals(n2 Namespace) bool {
--- a/share/namespace.go
+++ b/share/namespace.go
@@ -16,7 +16,7 @@ func (n Namespace) MarshalJSON() ([]byte, error) {
 	return json.Marshal(n.data)
 }
 
-// MarshalJSON decodes json bytes to the namespace.
+// UnmarshalJSON decodes json bytes to the namespace.
 func (n *Namespace) UnmarshalJSON(data []byte) error {
 	var buf []byte
 	if err := json.Unmarshal(data, &buf); err != nil {
--- a/share/namespace.go
+++ b/share/namespace.go
@@ -127,11 +127,11 @@ func (n Namespace) ValidateUserNamespace() error {
 
 // ValidateForData checks if the Namespace is of real/useful data.
 func (n Namespace) ValidateForData() error {
-	if err := n.ValidateUserNamespace(); err != nil {
-		return err
-	}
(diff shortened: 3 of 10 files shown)


share/share_test.go
```diff
@@ -238,3 +238,15 @@ func TestShareToBytesAndFromBytes(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, shares, reconstructedShares)
 }
+
+func TestMarshalShare(t *testing.T) {
+	sh := RandShares(1)
```
cristaloleg, line 243:
```suggestion
	sh := RandShares(1)[0]
```

So you don't need to keep `[0]` below.

share/blob.go
```diff
@@ -149,3 +150,20 @@ func SortBlobs(blobs []*Blob) {
 		return blobs[i].Compare(blobs[j]) < 0
 	})
 }
+
+// IsBlobNamespace returns a true if this namespace is a valid user-specifiable
+// blob namespace.
+func IsBlobNamespace(ns Namespace) bool {
```
cristaloleg, line 156:
```suggestion
func IsValidBlobNamespaceName(ns Namespace) bool {
```

WDYT ?

share/namespace.go
```diff
@@ -2,13 +2,35 @@ package share
...
 type Namespace struct {
 	data []byte
 }
 
+// MarshalJSON encodes namespace to the json encoded bytes.
+func (n Namespace) MarshalJSON() ([]byte, error) {
+	return json.Marshal(n.data)
+}
+
+// MarshalJSON decodes json bytes to the namespace.
```
rootulp, line 19:
```suggestion
// UnmarshalJSON decodes json bytes to the namespace.
```

share/namespace.go
```diff
@@ -203,6 +241,31 @@ func (n Namespace) Compare(n2 Namespace) int {
 	return bytes.Compare(n.data, n2.data)
 }
 
+// IsOutsideRange checks if the namespace is outside the min-max range of the given hashes.
+func (n Namespace) IsOutsideRange(leftHash, rightHash []byte) bool {
+	if len(leftHash) < NamespaceSize || len(rightHash) < 2*NamespaceSize {
```
rootulp, line 246: [question] how can the rightHash be 2 * the namespace size? I would expect leftHash and rightHash to be the same length and that length == NamespaceSize.

In this context, what is hash? I interpret it as the namespace of a left child or right child in an NMT.

share/random_namespace.go
```diff
@@ -38,22 +39,55 @@ func RandomBlobNamespace() Namespace {
...
 }
 
-// isBlobNamespace returns an true if this namespace is a valid user-specifiable
-// blob namespace.
-func isBlobNamespace(ns Namespace) bool {
-	if ns.IsReserved() {
-		return false
+// AddInt adds arbitrary int value to namespace, treating namespace as big-endian
+// implementation of int
+func AddInt(n Namespace, val int) (Namespace, error) {
```
rootulp, line 50: [question] why does celestia-node need to add ints to a namespace? It seems confusing that this library exposes a method to add an integer to a type (namespace) that doesn't really support that.

If you need a way to generate namespaces that are each one after the other, I propose we create a different utility method that generates and returns ascending namespaces instead of adding this function.

share/random_shares.go
```diff
@@ -0,0 +1,69 @@
+package share
+
+import (
+	"bytes"
+	"crypto/rand"
+	"fmt"
+	"sort"
+)
+
+// RandShares generate 'total' amount of shares filled with random data.
```
rootulp, line 10:
```suggestion
// RandShares generates total amount of shares and fills them with random data.
```

share/random_shares.go
```diff
@@ -0,0 +1,69 @@
...
+			panic(err)
+		}
+
+		shares[i] = *sh
+	}
+	sort.Slice(shares, func(i, j int) bool { return bytes.Compare(shares[i].ToBytes(), shares[j].ToBytes()) < 0 })
+	return shares
+}
+
+// RandSharesWithNamespace is same the as RandShares, but sets same namespace for all shares.
```
rootulp, line 38:
```suggestion
// RandSharesWithNamespace is the same as RandShares, but sets the same namespace for all shares.
```

rootulp (commented):
Not blocking but I have a few questions before approving.

share/random_namespace.go
vgonkivs: we need it for some tricky test cases in the node. I propose to explicitly add `t* testing.T` to the argument list

share/namespace.go
vgonkivs: This is the root hash of the NMT, and IIRC, it has a length of 90 bytes.

share/random_namespace.go
rootulp: Yea that works. Also could help to add a Godoc like this:

```go
// AddInt adds arbitrary int value to namespace, treating namespace as big-endian
// implementation of int. Note: should only be used for tests.
func AddInt(t *testing.T, n Namespace, val int) (Namespace, error) {
```

share/namespace.go
rootulp: thanks! Not blocking but the param names are confusing. It seems like `leftHash` is expected to be the length of a namespace and rightHash is expected to be the length of two namespaces concatenated. Would it makes sense to rename:

- leftHash => leftNamespace
- rightHash => rootHash

perhaps with additional comments that describe the expected length of both

vgonkivs: I'm starting to think that `IsOutsideRange` will be useful only for the node. Maybe it's better to keep it there as a helper function?

rootulp: Works for me

share/blob.go
vgonkivs: I'd just leave `IsValidBlobNamespace`

share/namespace.go
vgonkivs: moved to node

share/blob.go
```diff
@@ -149,3 +150,20 @@ func SortBlobs(blobs []*Blob) {
 		return blobs[i].Compare(blobs[j]) < 0
 	})
 }
+
+// IsValidBlobNamespace returns a true if this namespace is a valid user-specifiable
+// blob namespace.
+func IsValidBlobNamespace(ns Namespace) bool {
```
Wondertan, line 156: Seems like we have two patterns for namespace validators func signatures.
* ValidateForData
* IsValidBlobNamespace.

Worth unifying them.

[blocking]

share/namespace.go
```diff
@@ -2,13 +2,35 @@ package share
...
 	"fmt"
 )
 
 type Namespace struct {
 	data []byte
 }
 
+// MarshalJSON encodes namespace to the json encoded bytes.
+func (n Namespace) MarshalJSON() ([]byte, error) {
+	return json.Marshal(n.data)
```
Wondertan, line 16: We should consider making json marshalling human readable, like in hex instead.

Wondertan: Although, if that breaks something, then its not really worth it rn

share/namespace.go
```diff
@@ -104,6 +131,17 @@ func ValidateUserNamespace(version uint8, id []byte) error {
 	return validateID(version, id)
 }
 
+// ValidateForData checks if the Namespace is of real/useful data.
+func ValidateForData(n Namespace) error {
```
Wondertan, line 135: I wonder why we want to keep validators as functions rather than methods. Functions are harder to find when you work with namespaces, and usually, in the node, we want to validate them on the input. 

I guess it is more of a question to @cmwaters and @rootulp on why `ValidateUserNamespace` was made a func, while it could be `Namespace.ValidateUser` or `Namespace.IsUser`

[not blocking]

share/share.go
```diff
@@ -17,6 +18,20 @@ type Share struct {
...
+}
+
+func (s *Share) UnmarshalJSON(data []byte) error {
+	var buf []byte
+
+	if err := json.Unmarshal(data, &buf); err != nil {
+		return err
+	}
+	s.data = buf
+	return validateSize(s.data)
```
Wondertan, line 32: [blocking]

Check version

share/share.go
```diff
@@ -17,6 +18,20 @@ type Share struct {
 	data []byte
 }
 
+func (s Share) MarshalJSON() ([]byte, error) {
+	return json.Marshal(s.data)
+}
+
+func (s *Share) UnmarshalJSON(data []byte) error {
```
Wondertan, line 25: minor godoc for completeness

share/namespace.go
rootulp: Agreed methods seem preferable 👍 

Same [not blocking] comment, I don't think refactoring functions to methods needs to happen in this PR.

cmwaters: Yeah I agree with moving to methods

share/share.go
cmwaters: It depends what this gets used for. Erasure coded shares can have any random bytes of 512 in length. If we implemented this check here then if we were to unmarshal an EDS encoded in json it would always return an error

share/blob.go
cmwaters: Yeah `ValidateForData` seems redundant

share/share.go
Wondertan: Ok, good point. And there is no way to detect if that's a parity share by only looking at the bytes. Resolving

share/blob.go
vgonkivs: I think @Wondertan meant semantical unification.

share/blob_test.go
```diff
@@ -63,6 +63,16 @@ func TestBlobConstructor(t *testing.T) {
...
+
+	blob, err := NewBlob(ns, data, 0, nil)
+	require.NoError(t, err)
+	shares, err := blob.ToShares()
+	require.NoError(t, err)
+	blobList, err := parseSparseShares(shares)
+	require.NoError(t, err)
+	require.Len(t, blobList, 1)
+	require.Equal(t, blob, blobList[0])
+
```
rootulp, line 75:
```suggestion
```

share/blob.go
```diff
@@ -149,3 +150,30 @@ func SortBlobs(blobs []*Blob) {
...
+func (b *Blob) ToShares() ([]Share, error) {
+	splitter := NewSparseShareSplitter()
+	err := splitter.Write(b)
+	if err != nil {
+		return nil, err
+	}
+	return splitter.Export(), nil
+}
+
+// IsValidBlobNamespace returns a true if this namespace is a valid user-specifiable
```
rootulp, line 164:
```suggestion
// IsValidBlobNamespace returns true if this namespace is a valid user-specifiable
```

share/namespace.go
vgonkivs: Why should marshaling be human readable? This is a byte representation that will be sent over the wire. I've added `String()` method to the namespace, which makes it human-readable.

share/random_shares.go
```diff
@@ -0,0 +1,69 @@
...
+	"bytes"
+	"crypto/rand"
+	"fmt"
+	"sort"
+)
+
+// RandShares generates total amount of shares and fills them with random data.
+func RandShares(total int) []Share {
+	if total&(total-1) != 0 {
+		panic(fmt.Errorf("total must be power of 2: %d", total))
```
rootulp, line 13: [optional] return an error instead of `panic`ing. 

Tangent: it seems like a weird constraint that `RandShares` must be invoked with a power of `2`.

share/random_shares.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// RandSharesWithNamespace is the same as RandShares, but sets the same namespace for all shares.
+func RandSharesWithNamespace(namespace Namespace, namespacedAmount, total int) []Share {
+	if total&(total-1) != 0 {
+		panic(fmt.Errorf("total must be power of 2: %d", total))
+	}
+
+	if namespacedAmount > total {
+		panic(fmt.Errorf("withNamespace must be less than total: %d", total))
```
rootulp, line 45: [nit] this panic message is pretty cryptic. Proposal to return an error message like this instead:

```suggestion
		return []Share{], fmt.Errorf("namespacedAmount %v must be less than or equal to total: %v", namespacedAmount, total))
```

rootulp (dismissed):

share/random_shares.go
rootulp: If the refactor to return an error is done, then the rest of this function can also return errors instead of panicing

rootulp (dismissed):
Still LGTM 👍

share/namespace.go
```diff
@@ -2,21 +2,41 @@ package share
...
 // This should be used for user specified namespaces.
 func NewNamespace(version uint8, id []byte) (Namespace, error) {
-	if err := ValidateUserNamespace(version, id); err != nil {
-		return Namespace{}, err
-	}
-
-	return newNamespace(version, id), nil
+	ns := newNamespace(version, id)
+	return ns, ns.ValidateUserNamespace()
 }
```
cmwaters, line 40: Why have you changed this such that the namespace is still returned even if invalid?

vgonkivs: it will return an error anyway in case it happened. Decided to have fewer lines of code.

share/namespace.go
```diff
@@ -92,35 +108,51 @@ func (n Namespace) ID() []byte {
...
+	err := n.validateVersionSupported()
 	if err != nil {
 		return err
 	}
-	return validateID(version, id)
+	return n.validateID()
+}
+
+// ValidateForData checks if the Namespace is of real/useful data.
+func (n Namespace) ValidateForData() error {
```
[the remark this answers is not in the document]
vgonkivs: I will add `IsReserved`  as you syggested.

share/namespace.go
cmwaters: but it does allow someone to potentially ignore the error and use the invalid namespace

share/namespace.go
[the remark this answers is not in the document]
cmwaters: Is it possible to consolidate with `IsBlobNamespace` or is there a difference?

share/namespace.go
walldiss: Agree with @cmwaters, that returning both invalid value and error is unusual and might be confusing for user

share/random_blobs.go
```diff
@@ -0,0 +1,64 @@
+package share
```
walldiss, line 1: How about to put test funcs into share_test package instead of main one?

share/random_namespace.go
```diff
@@ -38,22 +41,53 @@ func RandomBlobNamespace() Namespace {
...
 }
 
-// isBlobNamespace returns an true if this namespace is a valid user-specifiable
-// blob namespace.
-func isBlobNamespace(ns Namespace) bool {
-	if ns.IsReserved() {
-		return false
+// AddInt adds arbitrary int value to namespace, treating namespace as big-endian
+// implementation of int. Note: should only be used for tests.
+func AddInt(t *testing.T, n Namespace, val int) Namespace {
```
walldiss, line 52: Why limit they use by tests only? It used to just return an error in node code

share/namespace.go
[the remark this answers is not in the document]
vgonkivs: we can use `ValidateForData` to check the share's namespace where it necessary. IsBlobNamespace contains an extra condition that checks `SupportedBlobNamespaceVersions`

share/namespace.go
vgonkivs: done

share/random_namespace.go
vgonkivs: I thought we need this function for testing purposes only. If not, could you please share more context

share/random_blobs.go
vgonkivs: there are some cases where we are accessing private fields of the structs(not blob actually)

share/random_namespace.go
walldiss: It is just a handlful function as namespace method and might be used by some users. Making it free func makes it much harder to find and generally does not allign with out package structure, where we make such functions as method on core type (namespace). I see it handful to be it as part of common API, even if it is not currently used by us.

vgonkivs: So, your proposal is to make it a method of Namespace structure, right?

share/namespace.go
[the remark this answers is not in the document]
cmwaters: Yeah I realise the extra `SupportedBlobNamespaceVersions` but can you perhaps tell me where you use the two? 

I would have thought we could just use `IsBlobNamespace`? 

When do you check the share's namespace? Could these shares also belong in the reserved namespace?

share/random_namespace.go
vgonkivs: moved to methods

walldiss (dismissed):

share/random_blobs.go
walldiss: Agreed to make to tackle it in separate PR

rootulp (dismissed):

share/namespace.go
Wondertan: I mean it's nice to keep human readable json human readable, but that's more of a general comment for everyone and a very minor concern, so resolving

share/namespace.go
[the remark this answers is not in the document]
Wondertan: There is slight difference. In node we had:
* `Validate`
* `ValidateForData`
* `ValidateForBlob`

`Validate` checks the correctness of the namespace and the actual validation we perform during its construction.  `Validate` is `ValidateUserNamespace` in this PR.

`ValidateForData` checks that the data namespace is for useful data, including reserved and user blob namespaces and excluding tail padding and parity. Reserved data are kept as they are still useful, and we want to allow the request of this data.

`ValidateForBlob` is on top of that to filter out reserved namespaces. This is needed to ensure valid user-only namespaces.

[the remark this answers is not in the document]
Wondertan: We can also consider privatising `Validate/ValidateUserNamespace` akin to Blobs as we do that validation in the Namespace constructor anyway

share/namespace.go
```diff
@@ -203,6 +255,52 @@ func (n Namespace) Compare(n2 Namespace) int {
 	return bytes.Compare(n.data, n2.data)
 }
 
+// AddInt adds arbitrary int value to namespace, treating namespace as big-endian
+// implementation of int. It could be helpful for users to create an adjacent namespaces.
```
renaynay, line 259:
```suggestion
// implementation of int. It could be helpful for users to create adjacent namespaces.
```

renaynay: We could also rename this method to indicate that's the usecase for it but i dont care that much

renaynay (dismissed):

share/namespace.go
```diff
@@ -92,35 +108,51 @@ func (n Namespace) ID() []byte {
...
+// ValidateForData checks if the Namespace is of real/useful data.
+func (n Namespace) ValidateForData() error {
+	if err := n.ValidateUserNamespace(); err != nil {
+		return err
+	}
+	if !n.IsUsableNamespace() {
+		return fmt.Errorf("invalid data namespace(%s): parity and tail padding namespace are forbidden", n)
+	}
+	return nil
 }
```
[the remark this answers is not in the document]
vgonkivs: We don't need to check for the Reserved here as it may break some cases.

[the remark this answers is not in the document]
Wondertan: The intention is to keep reserved data in. Reserved data is still valuable data of celestia transactions. We need a filter on namespace that allows requesting everything that can be requested excluding parity and tail padding. Please suggest a better name for the method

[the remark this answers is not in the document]
Wondertan: Basically including reserved filter here would break Sov Labs rejecting them when they request Celestias txs

rootulp (approved):
Still LGTM. Super optional but the PR title could be renamed to something more informative. Example: "feat: add missing utilities used by celestia-node"

share/namespace.go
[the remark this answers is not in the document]
cmwaters: Ok got it.

So we need both `ValidateForData` and `ValidateForBlob`

share/namespace.go
```diff
@@ -92,35 +116,60 @@ func (n Namespace) ID() []byte {
...
+	return hex.EncodeToString(n.data)
+}
+
 // ValidateUserNamespace returns an error if the provided version is not
 // supported or the provided id does not meet the requirements
 // for the provided version. This should be used for validating
 // user specified namespaces
-func ValidateUserNamespace(version uint8, id []byte) error {
-	err := validateVersionSupported(version)
+func (n Namespace) ValidateUserNamespace() error {
```
cmwaters, line 128: nit: can we just call this `Validate`. `ValidateUserNamespace` seems misleading because this also applies to namespaces that users can't use (like reserved namespaces or tailpadding)

share/namespace.go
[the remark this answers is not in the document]
cmwaters: Can this be removed and we just use `IsUsableNamespace`

share/namespace.go
```diff
@@ -92,35 +116,60 @@ func (n Namespace) ID() []byte {
...
+func (n Namespace) ValidateForBlob() error {
+	if err := n.ValidateForData(); err != nil {
+		return err
+	}
+
+	if !slices.Contains(SupportedBlobNamespaceVersions, n.Version()) {
+		return fmt.Errorf("blob version %d is not supported", n.Version())
+	}
+	return nil
 }
```
cmwaters, line 154: need to add the `IsReserved` check here as well.

Can we also add more of a description as to what constitutes `ValidForBlob`

cmwaters (changes requested):
Can we just get the validity rules for these methods right and then this seems good to merge

share/namespace.go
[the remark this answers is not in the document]
Wondertan: Its convenient to have a standartised error beijg returned. Otherwise users(we) have to make one evrytime we need this validity rule

share/namespace.go
Wondertan: Yes as per https://github.com/celestiaorg/go-square/pull/109#discussion_r1803362555

share/namespace.go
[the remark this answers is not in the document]
cmwaters: Ok, we can keep it if you guys want

share/namespace.go
vgonkivs: done

share/namespace.go
vgonkivs: renamed and made private

share/namespace.go
```diff
@@ -92,35 +116,68 @@ func (n Namespace) ID() []byte {
...
+		return fmt.Errorf("invalid data namespace(%s): parity and tail padding namespace are forbidden", n)
+	}
+	return nil
+}
+
+// ValidateForBlob verifies whether the Namespace is appropriate for blob data.
+// A valid blob namespace must meet two conditions: it cannot be reserved for special purposes,
+// and its version must be supported by the system. If either of these conditions is not met,
+// an error is returned indicating the issue. This ensures that only valid namespaces are
+// used when dealing with blob data.
```
Wondertan, line 148: optional: would be cool to have the same quality comment for ValidateForData

cmwaters (approved):
LGTM, Thanks!

share/namespace.go
vgonkivs: will do in a subsequent pr. Let's get it over the line

PR #113, opened 2024-10-07
chore(ci): add support for pushing proto definitions to Buf Schema Registry

rach-id:
since we have the blob definition in this repo, it would be good for it to be pushed to BSR to be used by downstream repos.


rootulp (approved):

cmwaters (approved):
utAck

PR #118, opened 2024-10-17
fix(inclusion): fix integer overflow+inefficient Round*PowerOfTwo

### LINKED ISSUE #117: inclusion, all: RoundDownPowerOfTwo and RoundUpPowerOfTwo are very inefficient and can cause running out of RAM due to an integer overflow
### Summary of Bug

While on a 3AM audit and vulnerability hunt, I noticed that Round*PowerOfTwo are heavily used within celestia's ecosytem by use of BlobMinSquareSize and SubTreeWidth.

If we examine the code for RoundUpPowerOfTwo, we can see this pattern https://github.com/celestiaorg/go-square/blob/8e9b275b5c6db9d82180ef2af01171fe12859d74/inclusion/blob_share_commitment_rules.go#L52-L59

which when read basically creates a value then per step of exponentiating 2 by 2 multiple times eventually will bump into a value less than the input, trying to get to that power of 2 in steps shall take a max of 64 loop iterations but sadly in that code there is an integer overflow when you pass in a signed integer `int*` due to this check
```go
        var result I = 1 
 	for result < input { 
 		result <<= 1 
 	} 
```
so when you have for example input=1<<64 - 1; that code starts with result=1 and in each step result is bit shifted to the left which is exponentiation for 2; but when we hit 1<<63; that value is still less than 1<<63 and in the next loop becomes 1<<64 and oops, maInt = 1<<64 - 1 hence that value overflows and becomes negative :-( -9223372036854775808 so that loop then continues and now the negative value when multiplied by 2 again warps around back to 0 but hey 0 << 1 == 0 and that loop iterates until you run out of RAM or your system kills it

## Remedy
We can make that code so much more efficient and not suspectible to any integer overflows using knowledge from bit shifting and fiddling by noticing a key component:
* a power of 2 has a single bit set in it e.g. 01000000 or 0000010 and if you subtracted by 1 any value with a bit mask of >= 0b10, the prior bit will be then 0b01 and fill the lower bits with 1 so for example 0b10000 - 0b01 = 0b01111 and from bit fiddling 0&1=0, 1&1=1 so the fast check for if a value is a power of 2 is simply if input&(input-1) == 0
* to find the next power of 2 for a non-power of 2, we just find the bit length and then left 1 by it so for example 6 = 0b110 which has 3 bits and 1<<3 = 0b1000 = 8 and for example 3 = 0b11 which has 2 bits and 1<<2 = 0b100 = 4 and that's why the logic works
* bits.Len64 uses a very efficient algorithm deBruijn with instrinsics, which I've attacked with no avail for these years and is well published of an algorithm for bit fiddlers and used in the Go standard library

```go
func RoundUpPowerOfTwo[I constraints.Integer](input I) I {
        if input <= 1 {
                return 1
        }
        if input&(input-1) == 0 {
                return input
        }
        return 1 << bits.Len64(uint64(input))
}

// RoundDownPowerOfTwo returns the next power of two less than or equal to input.
func RoundDownPowerOfTwo[I constraints.Integer](input I) (I, error) {
        if input <= 0 {
                return 0, fmt.Errorf("input %v must be positive", input)
        }
        if input&(input-1) == 0 {
                return input, nil
        }
        return 1 << (bits.Len64(uint64(input)) - 1), nil
}
```

which run in basically constant time and not susceptible to integer overflows and will always run in deterministic time and should actually have a positive impact on Celestia given that the callers for those functions are hot and the crux of Celestia

Kindly cc-ing @Wondertan @walldiss @liamsi @veorq

### Version

8e9b275b5c6db9d82180ef2af01171fe12859d74

### Steps to Reproduce

Run this code
```go
	RoundUpPowerOfTwo(math.MaxInt)
```

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

odeke-em:
This code fixes an integer flow using finer bit twiddling and even makes it much more efficient to always run deterministically in O(1) time essentially and not O(k) where k=log2(input) due to the prior k iterations that then caused the overflow when result became > maxInt.

While here also added some benchmarks and more test cases.

Fixes #117

### DIFF
--- a/inclusion/blob_share_commitment_rules.go
+++ b/inclusion/blob_share_commitment_rules.go
@@ -3,6 +3,7 @@ package inclusion
 import (
 	"fmt"
 	"math"
+	"math/bits"
 
 	"golang.org/x/exp/constraints"
 )
@@ -50,11 +51,19 @@ func RoundUpByMultipleOf(cursor, v int) int {
 
 // RoundUpPowerOfTwo returns the next power of two greater than or equal to input.
 func RoundUpPowerOfTwo[I constraints.Integer](input I) I {
-	var result I = 1
-	for result < input {
-		result <<= 1
+	if input <= 1 {
+		return 1
 	}
-	return result
+	if input&(input-1) == 0 { // It is already a power of 2
+		return input
+	}
+	var powUp I = 1 << bits.Len64(uint64(input))
+	if powUp < input {
+		// An overflow occurred due to a very large size
+		// of input and we should return a positive power of 2.
+		powUp = 1
+	}
+	return powUp
 }
 
 // RoundDownPowerOfTwo returns the next power of two less than or equal to input.
@@ -62,11 +71,12 @@ func RoundDownPowerOfTwo[I constraints.Integer](input I) (I, error) {
 	if input <= 0 {
 		return 0, fmt.Errorf("input %v must be positive", input)
 	}
-	roundedUp := RoundUpPowerOfTwo(input)
-	if roundedUp == input {
-		return roundedUp, nil
+	if input&(input-1) == 0 { // It is already a power of 2
+		return input, nil
 	}
-	return roundedUp / 2, nil
+
+	// Return 1 << (numberOfBits-1)
+	return 1 << (bits.Len64(uint64(input)) - 1), nil
 }
 
 // BlobMinSquareSize returns the minimum square size that can contain shareCount
@@ -79,6 +89,9 @@ func BlobMinSquareSize(shareCount int) int {
 // commitment over a given blob. The input should be the total number of shares
 // used by that blob. See ADR-013.
 func SubTreeWidth(shareCount, subtreeRootThreshold int) int {
+	if subtreeRootThreshold <= 0 {
+		return 1
+	}
 	// Per ADR-013, we use a predetermined threshold to determine width of sub
 	// trees used to create share commitments
 	s := (shareCount / subtreeRootThreshold)
--- a/inclusion/blob_share_commitment_rules_test.go
+++ b/inclusion/blob_share_commitment_rules_test.go
@@ -2,6 +2,7 @@ package inclusion_test
 
 import (
 	"fmt"
+	"math"
 	"testing"
 
 	"github.com/celestiaorg/go-square/v2/inclusion"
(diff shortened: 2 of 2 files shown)


inclusion/blob_share_commitment_rules.go
```diff
@@ -50,23 +51,32 @@ func RoundUpByMultipleOf(cursor, v int) int {
...
-	for result < input {
-		result <<= 1
+	if input <= 1 {
+		return 1
 	}
-	return result
+	if input&(input-1) == 0 { // It is already a power of 2
+		return input
+	}
+	var powUp I = 1 << bits.Len64(uint64(input))
```
rach-id, line 60: [Optional]
I would add a comment here explaining why this works

inclusion/blob_share_commitment_rules_test.go
```diff
@@ -326,86 +337,150 @@ func TestBlobMinSquareSize(t *testing.T) {
...
+	{
+		shareCount: defaultSubtreeRootThreshold - 1,
+		want:       1,
+	},
+	{
+		shareCount: defaultSubtreeRootThreshold * 2,
+		want:       2,
+	},
+	{
+		shareCount: (defaultSubtreeRootThreshold * 2) + 1,
```
rach-id, line 375: [Optional]
I would add something like:
```suggestion
		shareCount: (defaultSubtreeRootThreshold * 2) + 1, // 129
```
to avoid needing to calculate the numbers when reading the tests

rach-id (approved):
LGTM

inclusion/blob_share_commitment_rules_test.go
```diff
@@ -257,25 +258,35 @@ func TestRoundUpByMultipleOf(t *testing.T) {
...
+	{input: 4, want: 4},
+	{input: 5, want: 8},
+	{input: 8, want: 8},
+	{input: 11, want: 16},
+	{input: 511, want: 512},
+	{input: math.MaxInt32 - 1, want: 1 << 31},
+	{input: math.MaxInt32 + 1, want: 1 << 31},
+	{input: math.MaxInt32, want: 1 << 31},
+	{input: math.MaxInt >> 1, want: 1 << 62},
+	{input: math.MaxInt, want: 1},
```
rootulp, line 280: [blocking] this seems like unexpected behavior. I think we should return an error in this case

inclusion/blob_share_commitment_rules_test.go
```diff
@@ -257,25 +258,35 @@ func TestRoundUpByMultipleOf(t *testing.T) {
...
+}
```
rootulp, line 281: [blocking] these were previously defined inside the function `TestRoundUpPowerOfTwo`. Can you please move them back there so that their scope is confined to that function

inclusion/blob_share_commitment_rules_test.go
```diff
@@ -326,86 +337,150 @@ func TestBlobMinSquareSize(t *testing.T) {
...
+	},
+	{
+		shareCount: (defaultSubtreeRootThreshold * 5),
+		want:       8,
+	},
+	{
+		shareCount: (defaultSubtreeRootThreshold * defaultMaxSquareSize) - 1,
+		want:       128,
+	},
+}
```
rootulp, line 394: [blocking] please define `subtreeWidthTestCase` and `subtreeWidthTestCases` inside the function `TestSubTreeWidth` to confine their scope.

inclusion/blob_share_commitment_rules_test.go
```diff
@@ -326,86 +337,150 @@ func TestBlobMinSquareSize(t *testing.T) {
...
+	{input: 4, want: 4},
+	{input: 5, want: 4},
+	{input: 8, want: 8},
+	{input: 11, want: 8},
+	{input: 511, want: 256},
+	{input: math.MaxInt32 - 1, want: 1 << 30},
+	{input: math.MaxInt32, want: 1 << 30},
+	{input: math.MaxInt32 + 1, want: 1 << 31},
+	{input: math.MaxInt, want: 1 << 62},
+}
```
rootulp, line 475: Please define these inside `TestRoundDownPowerOfTwo`

inclusion/blob_share_commitment_rules_test.go
```diff
@@ -326,86 +337,150 @@ func TestBlobMinSquareSize(t *testing.T) {
...
 			assert.Equal(t, tc.want, got, i)
 		})
 	}
 }
 
-func TestRoundDownPowerOfTwo(t *testing.T) {
-	type testCase struct {
-		input int
-		want  int
+var sink any
```
rootulp, line 405: [question] should this be defined inside each Benchmark* test so that one benchmark does not impact another?

rootulp:
Closing b/c stale.

PR #119, opened 2024-10-19
test: add fuzzers for Builder.Export + inclusion.SubtreeWidth

odeke-em:
In a continuous bid to battle test and secure the network, this change adds 2 fuzzers for:
* Builder.Export
* inclusion.SubtreeWidth


rootulp:
golangci-lint is failing: https://github.com/celestiaorg/go-square/actions/runs/11416803480/job/31768417818?pr=119#step:5:37

cc: @odeke-em

odeke-em:
@rootulp thanks for the ping but I think that your repo's helpers are giving y'all bad tips and invalid suggestions
```go
  fuzz_test.go:15: File is not `gofumpt`-ed (gofumpt)
33
  	if err := os.MkdirAll(dirPath, 0755); err != nil {
```
is idiomatic code and the preferred way for quick throw away values in Go. I am going to change it but I highly suggest in the future taking a look at the suggestions.

cristaloleg:
> is idiomatic code and the preferred way for quick throw away values in Go. I am going to change it but I highly suggest in the future taking a look at the suggestions.

`gofumpt reports `0755` which should be `0o755`, not the problem you've mentioned.

odeke-em:
For the record, even that suggestion is incorrect per https://go.dev/play/p/CpKnMA6OAKj but let me push up the suggestion still.

odeke-em:
@cristaloleg we have a panic uncovered in 2 minutes of fuzzing per https://github.com/celestiaorg/go-square/issues/120. Kindly please help define what happens so that this change can then be merged. Thank you!

cristaloleg:
Maybe pedantic but definitely error-prone.

rootulp:
@odeke-em do you have a concrete suggestion on what linter settings you would like to be changed for this repo? https://golangci-lint.run/usage/linters/#gofumpt doesn't have a setting to disable just this rule:
 
Screenshot 2024-11-13 at 4 44 21 PM

so we would need to disable gofumpt entirely which doesn't seem desirable.

rootulp:
Closing b/c stale.

PR #155, opened 2025-06-12
fix: stop appending txs in `Build` after first one can not fit

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/4961
Part of https://github.com/celestiaorg/celestia-app/issues/4981
Unblocks https://github.com/celestiaorg/celestia-app/pull/4976

Pros
- This isn't consensus breaking because `square.Build` is only invoked in `PrepareProposal`.
- Stops proposers from generating a proposal that includes txs with out of order sequences. In other words, prevents the `account sequence mismatch` error we were hitting in ProcessProposal.

Cons
- Lowers the efficiency of squares. Previously square.Build was greedy and tried to pack as many txs as possible into a square. Now it stops as soon as the first tx can not fit. This means worse case that squares are 6 MiB full because the max tx size is 2 MiB and the max square size is 8 MiB.

### DIFF
--- a/square.go
+++ b/square.go
@@ -40,7 +40,10 @@ func Build(txs [][]byte, maxSquareSize, subtreeRootThreshold int) (Square, [][]b
 		}
 	}
 	square, err := builder.Export()
-	return square, append(normalTxs, blobTxs...), err
+	if err != nil {
+		return nil, nil, err
+	}
+	return square, append(normalTxs, blobTxs...), nil
 }
 
 // Construct takes the exact list of ordered transactions and constructs a square, validating that
--- a/square.go
+++ b/square.go
@@ -32,10 +32,16 @@ func Build(txs [][]byte, maxSquareSize, subtreeRootThreshold int) (Square, [][]b
 		if isBlobTx {
 			if builder.AppendBlobTx(blobTx) {
 				blobTxs = append(blobTxs, txBytes)
+			} else {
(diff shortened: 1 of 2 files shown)


square.go
```diff
@@ -32,15 +32,24 @@ func Build(txs [][]byte, maxSquareSize, subtreeRootThreshold int) (Square, [][]b
...
+			} else {
+				// The blobTx can not fit so exit the txs loop and return the square.
+				break
 			}
 		} else {
 			if builder.AppendTx(txBytes) {
 				normalTxs = append(normalTxs, txBytes)
+			} else {
+				// The normal tx can not fit so exit the txs loop and return the square.
+				break
```
evan-forbes, line 44: for posterity, leaving some thoughts here

this has the tradeoff of not including txs that otherwise would fit.

the max wasted space is that of the max transaction size, which is currently 2MiB

evan-forbes: irl, it would likely be less than the max size most of the time.

cmwaters (approved):
Can we create an issue that captures this change. Perhaps at a later point we can come up with a more sophisticated solution which maximizes square space without causing ordering errors

rootulp:
Closing in favor of 
- https://github.com/celestiaorg/go-square/pull/154
- https://github.com/celestiaorg/celestia-app/pull/4985

PR #164, opened 2025-07-16
test: ony the first share should parse the signer

rootulp:
Unit tests for https://github.com/celestiaorg/go-square/pull/163

### DIFF
--- a/share/parse_sparse_shares_test.go
+++ b/share/parse_sparse_shares_test.go
@@ -2,6 +2,7 @@ package share
 
 import (
 	"bytes"
+	"encoding/base64"
 	"fmt"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


share/parse_sparse_shares_test.go
```diff
@@ -143,15 +145,21 @@ func Test_parseSparseSharesWithNamespacedPadding(t *testing.T) {
 }
 
 func Test_parseShareVersionOne(t *testing.T) {
-	v1blob, err := NewV1Blob(MustNewV0Namespace(bytes.Repeat([]byte{1}, NamespaceVersionZeroIDSize)), []byte("data"), bytes.Repeat([]byte{1}, SignerSize))
```
rootulp, line 146: [note] this didn't previously repro the error because the blob data used was only `[]byte("data")` which fits in one share.

rootulp:
Closing b/c no longer needed. https://github.com/celestiaorg/go-square/pull/163 includes unit tests that reproduce the error. We don't need the exact blob from the original issue.

Issue #143, opened 2025-03-28
Upgrade to golangci-lint v2


rootulp:
## Context

https://github.com/celestiaorg/go-square/pull/142

## Problem

That dependabot PR bumped the action but it didn't bump golangci-lint to v2

## Proposal

1. Upgrade to golangci-lint v2: https://github.com/golangci/golangci-lint/releases/tag/v2.0.2
2. Upgrade the action to v7: https://github.com/golangci/golangci-lint-action/releases/tag/v7.0.0

PR #172, opened 2025-08-22
chore: upgrade to golangci-lint v2

rootulp:
Closes https://github.com/celestiaorg/go-square/issues/143


.golangci.yml
```diff
@@ -1,20 +1,35 @@
+version: "2"
```
rootulp, line 1: this file is the output of `golangci-lint migrate`

Issue #184, opened 2025-09-15
Test that verifies go-square v2.3.2 was breaking


rootulp:
## Context

@renaynay and @evan-forbes debugged last week and identified that go-square v2.3.2 was consensus breaking. Specifically on Arabica height [8122437](https://rpc.celestia-arabica-11.com/block?height=8122437)

## Problem

The breaking line is likely: 
https://github.com/celestiaorg/go-square/pull/171/files#diff-5dce70063c4b1b07e611b3fcd32da0445a0b95970028c8992beb9c955e38320fR437

## Proposal

1. Add a unit test that builds a square with that block. Verify that the square produced with go-square v2.3.2 does not match v2.3.1
2. If v2.3.2 was breaking, consider redacting it
3. Create go-square v3.0.0 so that celestia-node can use go-square v2.3.1 for celestia-app app version 5 blocks and go-square v3 for celestia-app app version 6 blocks.

rootulp:
I verified that the Arabica block in question produces two different original data squares.
1. v2.3.1 produces a hash of `3c76056375f8720a04d543514f4b4cdd8a9c59ab8ec57dabf11385d117e79f71`
2. v2.3.2 produces a hash of `1880b0e77b4684cb0c0b331be3c9f99f1507933e05a1352cdbaabab34e8fc03f`

PR #186, opened 2025-09-15
test: verify go-square v2.3.2 breaking change

rootulp:
Closes https://github.com/celestiaorg/go-square/issues/184


rach-id (approved):

rootulp:
> Is there a more general way of catching future regressions?

One idea is to include more historical blocks in this type of regression test. This approach isn't "general" but does expand the coverage of the regression test.

Issue #188, opened 2025-09-15
auto-request-review fails


rootulp:
## Context

https://github.com/celestiaorg/go-square/actions/runs/17745479120/job/50429581126?pr=182

## Problem

Auto request reviewers CI keeps failing

## Proposal

Remove it

PR #189, opened 2025-09-15
chore: remove auto-request-review automation

rootulp:
Closes https://github.com/celestiaorg/go-square/issues/188


evan-forbes (approved):
why remove? just that it fails?

rach-id (approved):
we should find a better way to add reviewers to PRs

rootulp:
Even without this CI workflow, reviewers get auto-added so seems unnecessary to me.

Issue #116, opened 2024-10-16
share: NewShare should return value instead of pointer


Wondertan:
There is literally no value in returning a pointer in this case. The Share struct is smaller and has only one ptr field `data` , leading to copying the value instead of pointer indirection having better performance. Also, working with a pointer is slightly more annoying. 

Refs:
* https://dev.to/anubhav023/go-benchmarks-does-pass-by-pointer-really-make-a-difference-1540
* https://blog.devtrovert.com/p/go-ep3-pass-values-not-pointers

PR #190, opened 2025-09-15
refactor!: NewShare returns value instead of pointer

rootulp:
Closes https://github.com/celestiaorg/go-square/issues/116

Motivation: I plan on creating a v3.0.0 from `main` soon so I'm trying to merge and this seemed like a small breaking PR we could include. I'm not strongly opinionated on if we want/need to block v3.0.0 on it.

### DIFF
--- a/share/random_shares.go
+++ b/share/random_shares.go
@@ -21,15 +21,15 @@ func RandShares(total int) ([]Share, error) {
 			panic(err)
 		}
 
-		sh, err := NewShare(shr)
+		share, err := NewShare(shr)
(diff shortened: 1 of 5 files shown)


evan-forbes (approved):
approving to :shipit: 

node team raised a good point that breaking the apis complicate dynamically switching at an upgrade height, so if that becomes an issue then in order to optimize for getting v6 out imo we should just cut a v4 that focuses on api breaks but is consensus backwards compatible with v3

cmwaters (approved):

PR #192, opened 2025-09-16
chore: prep for v3 release

rootulp:
Prepare for creating the v3.0.0 release by changing the go `module` declaration to have a `/v3` suffix


cmwaters (approved):
There's a few cases where it seems there are unnecessary aliasing of the imports (which I'm surprised a linter doesn't complain about)

PR #194, opened 2025-09-17
chore!: rename and regenerate blob package

cmwaters:
This is intended to prevent the proto registry conflicts that occur when trying to import both v2 and v3 of go-square

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -7,7 +7,7 @@ import (
 	"sort"
 
 	"github.com/celestiaorg/go-square/v3/inclusion"
-	v1 "github.com/celestiaorg/go-square/v3/proto/blob/v1"
+	v2 "github.com/celestiaorg/go-square/v3/proto/blob/v2"
 	"github.com/celestiaorg/go-square/v3/share"
 	"github.com/celestiaorg/go-square/v3/tx"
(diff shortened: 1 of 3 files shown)


proto/blob/v2/blob.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
-package proto.blob.v1;
+package proto.blob.v2;
```
rootulp, line 2: [optional] if we have to increment this for every future go-release version because node will import the latest version and the trailing version, then maybe this version can match the go-square version. In other words

```suggestion
package proto.blob.v3;
```

rootulp (approved):
LGTM

evan-forbes (approved):
~~I'm curious if anyone used v2 or what the diff was there that justified a new version in the first place?~~
edit
just to clarify for future reference:

our current strategy is to delete the old version of the protos, then create a version every major release?

proto/blob/v2/blob.proto
cmwaters: Yeah I thought about that afterwards and then didn't get around to implementing. It would be a small consolation to have it match the square version - just to keep something simpler - but perhaps we can do that the next major release

PR #202, opened 2025-11-08
feat: append pay for fibre txs

rootulp:
Targets feature/fibre

Part of to writing MsgPayForFibre data to the data square

PR #205, opened 2025-11-11
feat: append system level blob

rootulp:
Blocked on https://github.com/celestiaorg/go-square/pull/202


evan-forbes (approved):
out of curiosity, the system blobs should be constant size, correct? what is the size / worst case size?

PR #212, opened 2025-11-25
chore: prep for v4 release

rootulp:
Bump the go module to v4 so that we can use an RC created from the feature/fibre branch.


rootulp:
I think the buf failure is expected:

> Error: File option "go_package" changed from "github.com/celestiaorg/go-square/v3/proto/blob/v2" to "github.com/celestiaorg/go-square/v4/proto/blob/v2".

evan-forbes (approved):

PR #266, opened 2026-06-11
fix!: domain-separate and length-prefix Blob.Hash preimage

rootulp:
## Overview

Resolves the open security advisory [GHSA-jv5p-xwcc-q8mf](https://github.com/celestiaorg/go-square/security/advisories/GHSA-jv5p-xwcc-q8mf) by hardening the `Blob.Hash()` preimage construction. Details are tracked in the advisory and the Linear issue.

`Hash()` now prefixes the preimage with a domain tag and length-prefixes every variable-length field, so distinct blobs can no longer be crafted to produce the same hash.

## Testing

Added a regression test (`TestBlobHash/cross-version_collision_is_not_possible`) that fails on the old implementation and passes with the fix. Full suite and race tests pass.

## Breaking change

`Hash()` output changes for all blobs. Downstream callers that persist or key on `Hash()` output (mempool dedup, indexer caches, etc.) must re-compute it.

Closes PROTOCO-1760

### DIFF
--- a/share/blob.go
+++ b/share/blob.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"hash"
 	"sort"
 
 	v4 "github.com/celestiaorg/go-square/v4/proto/blob/v4"
@@ -159,14 +160,30 @@ func NewBlobFromProto(pb *v4.BlobProto) (*Blob, error) {
 	)
 }
 
-// Hash returns a SHA-256 hash of all the blob's fields.
+// blobHashDomain is a domain separation tag prefixed to the Hash preimage so
+// that the hash can not be confused with hashes computed over other data.
+const blobHashDomain = "celestia.blob.v1\x00"
+
+// Hash returns a SHA-256 hash of all the blob's fields. Each variable-length
+// field is length-prefixed so that distinct blobs can not be made to produce
+// the same preimage (e.g. a v0 blob whose data embeds a v1 signer).
 func (b *Blob) Hash() []byte {
 	h := sha256.New()
-	h.Write(b.Namespace().Bytes())
-	h.Write(b.Data())
+	h.Write([]byte(blobHashDomain))
+	writeLengthPrefixed(h, b.Namespace().Bytes())
+	writeLengthPrefixed(h, b.Data())
 	h.Write([]byte{b.ShareVersion()})
-	h.Write(b.Signer())
+	writeLengthPrefixed(h, b.Signer())
 	return h.Sum(nil)
+}
+
+// writeLengthPrefixed writes a 4-byte big-endian length prefix followed by the
+// data itself to h.
+func writeLengthPrefixed(h hash.Hash, data []byte) {
+	var prefix [4]byte
+	binary.BigEndian.PutUint32(prefix[:], uint32(len(data)))
+	h.Write(prefix[:])
+	h.Write(data)
 }
 
 // Namespace returns the namespace of the blob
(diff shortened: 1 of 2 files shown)


rootulp:
## Why this should land as-is (in-place fix, no new method / deprecation)

I analyzed the downstream impact before deciding on the approach. Conclusion: the in-place change is correct and the "breaking change" is theoretical.

**No consumer persists or keys on `Blob.Hash()` output.** Grepped celestia-node, celestia-app, celestia-app-fibre, and celestia-indexer — there are **zero production callers** of `share.Blob.Hash()`, directly or via celestia-node's `blob.Blob` wrapper (which embeds `*share.Blob` but never invokes the promoted method). Every `.Hash()` hit in those repos is on `coretypes.Tx`, block/header types, the DAH, or merkle roots — unrelated.

**`Blob.Hash()` only exists in v4, and there is no stable v4 release.** go-square tags top out at `v4.0.0-rc4`; consumers are pinned to release candidates. So this method has never been part of a stable, released API contract, and changing it now breaks nothing in practice. The only action item is to ensure this lands **before** the stable `v4.0.0` tag.

**Why not a new method + deprecate the old one?** Deprecation protects existing consumers of a released, stable API — there are none here. Keeping the old colliding `Hash()` alive as a "compatible" path would just preserve the vulnerability for anyone who keeps calling it, and a `HashV2()`-style split would invite integrators to pick the insecure one. The old behavior is a security bug, so the right move is to fix it in place.

The `BREAKING CHANGE` footer is kept for honesty in the RC-to-RC changelog, but no stable release exposes this and no known consumer persists the output.

ninabarbakadze (commented):
I can't open the linked security advisory and I'd like to read it before going through the PR 🙏🏻

rootulp:
Good call. Just added @ninabarbakadze and @rach-id to have access to the security advisory.

share/blob.go
```diff
@@ -159,16 +160,32 @@ func NewBlobFromProto(pb *v4.BlobProto) (*Blob, error) {
...
+	writeLengthPrefixed(h, b.Data())
 	h.Write([]byte{b.ShareVersion()})
-	h.Write(b.Signer())
+	writeLengthPrefixed(h, b.Signer())
 	return h.Sum(nil)
 }
 
+// writeLengthPrefixed writes a 4-byte big-endian length prefix followed by the
+// data itself to h.
+func writeLengthPrefixed(h hash.Hash, data []byte) {
```
rach-id, line 184: why do we need this? if I understand correctly, introducing separators between each segment of the hashed data is enough, no?

If we don't want to do that, we could use the length and a custom separator `<length>:data` for example, as separators.

Using both gives us no extra security as far as I know

rootulp: Separators alone are not enough here because `data` and `signer` are arbitrary bytes, so a field can contain the separator itself. E.g. with separator `|`:

- `data = "a|b"`, `signer = "c"`
- `data = "a"`, `signer = "b|c"`

both produce `...a|b|c...` — two distinct blobs, same preimage. A fixed-width length prefix makes the encoding unambiguous because a parser always knows exactly where each field ends.

`<length>:data` is effectively what this implementation does: once the length is a fixed-width prefix, the boundary is already unambiguous, so the `:` separator adds nothing. Agreed that combining lengths with separators buys no extra security — that is why there is no separator, just `<length><data>`.

The domain tag serves a different purpose than the length prefixes: it prevents `Blob.Hash()` from colliding with a SHA-256 hash computed over the same bytes in some other context, while the length prefixes prevent field-boundary ambiguity within blob hashing. Added a note explaining this to `writeLengthPrefixed` in d16e977.

rach-id (approved):

PR #274, opened 2026-08-13
fix: resolve duplicate tx body_bytes to the last occurrence

rootulp:
## Description

`TryParseFibreTx` decided whether transaction bytes are a fibre tx by reading the body that protobuf-go produces for `cosmos.tx.v1beta1.Tx`. In that schema field 1 is the singular message `Tx.body`, and protobuf-go merges repeated occurrences of a message field, concatenating the inner `repeated` messages. The Cosmos SDK reads the same wire field as `TxRaw.body_bytes`, a scalar `bytes` field, where gogoproto keeps only the last occurrence.

So when field 1 is repeated, the two disagree about which message comes first. go-square has to follow the SDK, since the SDK is what decoded, validated and signed over the transaction.

This resolves field 1 to its last top-level occurrence before classifying. Transactions from standard signers encode field 1 exactly once, so their classification is unchanged and the added wire scan is the only cost.

Also guards `AppendFibreTx` against a nil `FibreTx` or nil system blob. It dereferenced `fibreTx.SystemBlob` on its first line, so a caller handing it the nil that `TryParseFibreTx` returns for a non-fibre tx got a panic rather than an error.

Details in PROTOCO-2288.

## Testing

- `TestTryParseFibreTxDuplicateBodyField` covers both orderings of a repeated field 1, a duplicated identical body, and a last occurrence that is not a valid `TxBody`. Both orderings fail without the change.
- `TestBuilderAppendFibreTxNil` covers the nil guard; it panics without the change.
- `go test -race ./...` passes.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -185,6 +185,9 @@ func (b *Builder) RevertLastBlobTx() error {
 // adds the raw SDK tx bytes to PayForFibreTxs (compact shares) and the system
 // blob to Blobs (sparse shares). It returns false if there is not enough space.
 func (b *Builder) AppendFibreTx(fibreTx *tx.FibreTx) (bool, error) {
+	if fibreTx == nil || fibreTx.SystemBlob == nil {
+		return false, errors.New("fibre tx and its system blob must be non-nil")
+	}
 	if fibreTx.SystemBlob.ShareVersion() != share.ShareVersionTwo {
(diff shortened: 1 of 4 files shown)


rootulp:
Duplicate of #275, which carries the same fix (nil guard + last-occurrence body resolution) with the more accurate `fix!:` marking. Closing in favour of that one; the one test case unique to this PR is noted at https://github.com/celestiaorg/go-square/pull/275#issuecomment-5276028568.

PR #275, opened 2026-08-13
fix!: decode duplicate tx body field the way the SDK does

rootulp:
I plan on releasing this and backporting it to celestia-app v9.x.

Decode the transaction body field the way the Cosmos SDK does. The SDK reads that wire field as `TxRaw.body_bytes`, a scalar `bytes`, and keeps the last occurrence; protobuf-go merges every occurrence of a singular message field. go-square now follows the SDK.

Transactions carrying the field exactly once — every transaction a standard signer produces — decode exactly as before. Also guards `AppendFibreTx` against a nil fibre tx or system blob.

Note for reviewers: breaking, because it changes the square produced for transactions that carry the field more than once.

Details: https://linear.app/celestia/issue/PROTOCO-2288

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -185,6 +185,9 @@ func (b *Builder) RevertLastBlobTx() error {
 // adds the raw SDK tx bytes to PayForFibreTxs (compact shares) and the system
 // blob to Blobs (sparse shares). It returns false if there is not enough space.
 func (b *Builder) AppendFibreTx(fibreTx *tx.FibreTx) (bool, error) {
+	if fibreTx == nil || fibreTx.SystemBlob == nil {
+		return false, fmt.Errorf("fibre tx and its system blob must be non-nil")
+	}
 	if fibreTx.SystemBlob.ShareVersion() != share.ShareVersionTwo {
 		return false, fmt.Errorf("system blobs must use ShareVersionTwo, got version %d", fibreTx.SystemBlob.ShareVersion())
 	}
--- a/builder_test.go
+++ b/builder_test.go
@@ -1561,3 +1561,16 @@ func TestBlobShareRangeNotSupportedForSystemBlobs(t *testing.T) {
 	ns2Range := share.GetShareRangeForNamespace(sq, ns2)
 	assert.False(t, ns2Range.IsEmpty(), "system blob namespace ns2 should be present in the square")
 }
+
+// TestBuilderAppendFibreTxNil asserts AppendFibreTx reports an error rather than
+// panicking when handed a nil fibre tx.
+func TestBuilderAppendFibreTxNil(t *testing.T) {
+	builder, err := square.NewBuilder(8, 64)
+	require.NoError(t, err)
+
+	added, err := builder.AppendFibreTx(nil)
+	require.Error(t, err)
+	require.False(t, added)
+	require.Empty(t, builder.PayForFibreTxs)
+	require.Empty(t, builder.Blobs)
+}
--- a/tx/pay_for_fibre.go
+++ b/tx/pay_for_fibre.go
@@ -7,6 +7,7 @@ import (
 	cosmostx "github.com/celestiaorg/go-square/v4/proto/cosmos/tx/v1beta1"
 	"github.com/celestiaorg/go-square/v4/share"
 	"github.com/cosmos/btcutil/bech32"
+	"google.golang.org/protobuf/encoding/protowire"
 	"google.golang.org/protobuf/proto"
 )
 
@@ -27,11 +28,15 @@ func TryParseFibreTx(txBytes []byte) (*FibreTx, error) {
 	if err := proto.Unmarshal(txBytes, &sdkTx); err != nil {
 		return nil, nil
 	}
-	if sdkTx.Body == nil || len(sdkTx.Body.Messages) == 0 {
+	body, err := authoritativeBody(txBytes, &sdkTx)
+	if err != nil {
+		return nil, nil
+	}
+	if body == nil || len(body.Messages) == 0 {
 		return nil, nil
 	}
 
-	anyMsg := sdkTx.Body.Messages[0]
+	anyMsg := body.Messages[0]
 	if anyMsg.TypeUrl != MsgPayForFibreTypeURL {
 		return nil, nil
 	}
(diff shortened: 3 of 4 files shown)


rootulp:
Consolidating #274 into this PR — closing that one as a duplicate. The two were the same fix (nil guard + last-occurrence resolution); this one is kept because `fix!:` is the more accurate marking, since the square produced for duplicate-field transactions changes.

One case from #274 worth porting into `TestTryParseFibreTxDuplicateBody`: a last occurrence that is **not a valid `TxBody`**. It exercises the `proto.Unmarshal` error path in `authoritativeBody`, which nothing currently covers.

```go
{
    name:      "invalid trailing body is not a fibre tx",
    bodies:    [][]byte{fibreBody, {0xFF}},
    wantFibre: false,
},
```

Worth having because it pins the intended behavior for that path: the SDK's decoder rejects those bytes outright, so treating them as "not a fibre tx" keeps go-square aligned with it rather than falling back to the merged body.

rach-id (commented):
Here are a couple approaches that can be explored if not already:
```
  ## 1. Generate a local TxRaw model — recommended

  The fundamental problem is not protobuf parsing itself. It is that go-square currently decodes wire field 1 using the wrong schema:

  message Tx {
    TxBody body = 1;
  }

  The SDK decodes those same transaction bytes as:

  message TxRaw {
    bytes body_bytes = 1;
    bytes auth_info_bytes = 2;
    repeated bytes signatures = 3;
  }

  For duplicate scalar fields, the last value replaces earlier values. Duplicate embedded messages are merged. That distinction is part of protobuf semantics, so the safest fix is to give the protobuf runtime
  the correct schema and let it do the work. See the SDK’s canonical TxRaw schema (https://github.com/celestiaorg/cosmos-sdk/blob/v0.52.8/proto/cosmos/tx/v1beta1/tx.proto#L29-L47) and the protobuf
  last-one-wins rules (https://protobuf.dev/programming-guides/encoding/#last-one-wins).

  Add TxRaw to go-square’s existing minimal Cosmos schema:

  message TxRaw {
    bytes body_bytes = 1;
    bytes auth_info_bytes = 2;
    repeated bytes signatures = 3;
  }

  // Retain this for source compatibility, but stop using it for wire txs.
  message Tx {
    option deprecated = true;
    TxBody body = 1;
  }

  Then parsing becomes:

  func decodeTxBody(txBytes []byte) (*cosmostx.TxBody, error) {
  	var raw cosmostx.TxRaw
  	if err := proto.Unmarshal(txBytes, &raw); err != nil {
  		return nil, err
  	}

  	var body cosmostx.TxBody
  	if err := proto.Unmarshal(raw.BodyBytes, &body); err != nil {
  		return nil, err
  	}
  	return &body, nil
  }

  This exactly fixes the case from the review:

  body_bytes = ff
  body_bytes = <valid Fibre body>

  The first value is treated as opaque bytes and overwritten. Only the authoritative final value is parsed as TxBody.

  The inverse also behaves correctly:

  body_bytes = <valid Fibre body>
  body_bytes = ff

  The final value is retained, then parsing it fails—matching the SDK.

  ### Why this is safest

  - There is no handwritten tag, length, group, or unknown-field handling.
  - Future protobuf-runtime fixes automatically apply.
  - The code directly expresses SDK decoding: outer TxRaw, then inner TxBody.
  - Invalid earlier bodies do not get parsed.
  - No Cosmos SDK dependency is added.
  - The existing Tx type can remain deprecated, making the schema change additive.

  This approach already exists in closed PR #276 (https://github.com/celestiaorg/go-square/pull/276). It was closed in favor of #275, not because of a technical flaw. I ran an additional
  malformed-earlier/valid-final regression against that branch; it passes.

  I would reuse the core of #276 but omit its unrelated ValidateADR027TxRaw helper. That helper manually parses wire data and is not needed for this targeted correction.

  ### Suggested test matrix

  At minimum:

  - One valid Fibre body → Fibre.
  - Fibre then normal → normal.
  - Normal then Fibre → Fibre.
  - Malformed earlier then valid Fibre → Fibre.
  - Valid Fibre then malformed final → not Fibre.
  - Several malformed/valid bodies with a valid final body → final body wins.
  - Duplicate body fields followed by real auth-info and signature fields.
  - One parity test in celestia-app that feeds the same vectors through the actual SDK decoder and go-square classifier.

  The last test belongs in celestia-app, avoiding an SDK dependency in go-square.

  ## 2. Use wrapperspb.BytesValue as the envelope

  There is a very small alternative requiring no new .proto or generated code.

  google.protobuf.BytesValue has this wire shape:

  message BytesValue {
    bytes value = 1;
  }

  Its field 1 is wire-compatible with TxRaw.body_bytes. Fields 2 and 3 become discarded unknown fields.

  var rawBody wrapperspb.BytesValue
  if err := (proto.UnmarshalOptions{
  	DiscardUnknown: true,
  }).Unmarshal(txBytes, &rawBody); err != nil {
  	return nil, nil
  }

  var body cosmostx.TxBody
  if err := proto.Unmarshal(rawBody.Value, &body); err != nil {
  	return nil, nil
  }

  I prototyped this against PR #275, added the malformed-earlier/valid-final regression, and ran go test ./...; everything passed.

  ### Advantages

  - Deletes the entire manual authoritativeBody scanner.
  - No new generated files.
  - No new module dependency.
  - Very easy to backport.
  - The protobuf runtime provides the required last-one-wins behavior.

  ### Disadvantages

  It is type-punning: transaction bytes are not semantically a BytesValue; they merely have a compatible field 1. That is correct at the wire level but less obvious to future maintainers.

  It also deliberately ignores the types of fields 2 and 3. That is acceptable if TryParseFibreTx is only classifying transactions the application’s SDK decoder already accepted, but it is not a complete TxRaw
  validation mechanism.

  I would use this only if minimizing the backport diff is more important than making the schema explicit. It needs a strong comment explaining why BytesValue is used and a permanent malformed-earlier
  regression test.
```

I would prefer an intermediate type or using existing proto functionality rather than writing proto scaning code that no one understands how to maintain and is more error prone. 

Let's make writing manual proto parsers the last resort

rootulp:
I tried your first alternative in https://github.com/celestiaorg/go-square/pull/276 and closed it because it's a larger diff than this PR. I'll re-open and close this.

PR #276, opened 2026-08-13
fix!: decode transactions through TxRaw, as the SDK does

rootulp:
Alternative to #275, which resolves the same issue by scanning the wire format with `protowire`. This declares `TxRaw` and decodes in the SDK's two steps — bytes to `TxRaw`, then `body_bytes` to `TxBody` — so duplicate-field resolution falls out of protobuf semantics instead of being patched in for one field. `Tx` is deprecated rather than deleted because `buf breaking` guards the published module.

Also adds `tx.ValidateADR027TxRaw`, a port of the SDK's ADR-027 check. It is exported and deliberately not called during square construction. Carries over the `AppendFibreTx` nil guard from #275.

Note for reviewers: breaking, because it changes the square produced for transactions that carry the field more than once.

Details: https://linear.app/celestia/issue/PROTOCO-2288


rootulp:
Closing in favor of #275

tx/adr027.go
rach-id, line 1: wowowowo this is already crazy complex 😆

PR #277, opened 2026-08-13
feat!: require callers to classify transactions

rootulp:
Closes https://linear.app/celestia/issue/PROTOCO-2354

Construct, NewBuilder, TxShareRange, and BlobShareRange now take pre-classified transactions, and TryParseFibreTx along with the duplicated `cosmos/tx` and `celestia/fibre` protobuf schemas are removed. go-square decodes formats it owns (BlobTx, IndexWrapper, Blob) and no longer decodes formats the SDK owns, so it cannot reach a different answer than celestia-app about what a transaction is. The application passes the synthesized system blob in.

Classifications are validated at the boundary: empty bytes, inconsistent fibre classifications, and a BlobTx classified as fibre are rejected. Zero-length transactions were previously accepted as normal txs and now error; see COMPATIBILITY.md.

The deprecated `Build` is removed in the same breaking release.

Note for reviewers: consensus-affecting and breaking. celestia-app needs its companion change to compile against this, and both must land before v10 ships.


rootulp:
Draft PR until https://github.com/celestiaorg/go-square/pull/275 is merged and released.

classified_tx.go
```diff
@@ -0,0 +1,77 @@
...
+// transaction is a MsgPayForFibre transaction requires the SDK's schema and the
+// application's rules, so the caller makes that decision and passes the result
+// in.
+//
+// This is not a stylistic preference. When go-square classified transactions
+// itself it used a different protobuf schema than the SDK, and the two reached
+// different answers about the same bytes: one square from the proposer, another
+// from every validator re-deriving it. A library that re-derives a
+// consensus-critical decision its caller has already made can only agree with
+// the caller by coincidence.
```
rootulp: This can be removed

rach-id (dismissed):
👍

rootulp:
Already got Rachid's approval, after that I just resolved a merge conflict.

